PR #58, opened 2018-06-20
Make article model extendable

mmachatschek:
This makes the plugin extendable by adding a service (the articlefactory) which can be overriden via a services in other plugins

### DIFF
--- a/FinSearchAPI/BusinessLogic/FindologicArticleFactory.php
+++ b/FinSearchAPI/BusinessLogic/FindologicArticleFactory.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace FinSearchAPI\BusinessLogic;
+
+use FinSearchAPI\BusinessLogic\Models\FindologicArticleModel;
+use Shopware\Models\Article\Article;
+
+class FindologicArticleFactory
+{
+    /**
+     * FindologicArticleModel constructor.
+     *
+     * @param Article $shopwareArticle
+     * @param string  $shopKey
+     * @param array   $allUserGroups
+     * @param array   $salesFrequency
+     *
+     * @throws \Exception
+     */
+    public function createFindologicArticle(Article $shopwareArticle, $shopKey, array $allUserGroups, array $salesFrequency)
+    {
+        return new FindologicArticleModel($shopwareArticle, $shopKey, $allUserGroups, $salesFrequency);
+    }
+}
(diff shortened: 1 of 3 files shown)


FinSearchAPI/ShopwareProcess.php
```diff
@@ -91,6 +90,8 @@ public function getAllProductsAsXmlArray($start = 0, $count = 0)
         /** @var array $allUserGroups */
         $allUserGroups = $this->customerRepository->getCustomerGroupsQuery()->getResult();
 
+        $findologicArticleFactory = Shopware()->Container()->get('fin_search_api.article_model_factory');;
```
ihanli: Remove the second `;`

FinSearchAPI/Resources/services.xml
```diff
@@ -23,6 +23,10 @@
 
         <service id="fin_search_api.shopware_process" class="FinSearchAPI\ShopwareProcess">
 
+        </service>
+
+        <service id="fin_search_api.article_model_factory" class="FinSearchAPI\BusinessLogic\FindologicArticleFactory">
```
ihanli: Isn't it possible to use `FindologicArticleModel` directly?

FinSearchAPI/ShopwareProcess.php
mmachatschek: Done

FinSearchAPI/Resources/services.xml
mmachatschek: I guess not because in this case we need to return new Instances of the `FindologicArticleModel`. See here https://developers.shopware.com/developers-guide/services/#calling-the-service

FinSearchAPI/BusinessLogic/FindologicArticleFactory.php
```diff
@@ -0,0 +1,24 @@
...
+     * FindologicArticleModel constructor.
+     *
+     * @param Article $shopwareArticle
+     * @param string  $shopKey
+     * @param array   $allUserGroups
+     * @param array   $salesFrequency
+     *
+     * @throws \Exception
+     */
+    public function createFindologicArticle(Article $shopwareArticle, $shopKey, array $allUserGroups, array $salesFrequency)
```
ihanli: Since all the info regarding FINDOLOGIC and articles is already in the class name, I'd propose to rename this method to just `create`. So you will end up with something like
```
$findologicArticleFactory->create(...);
```

mmachatschek: Done. Also updated the docblock info about the method

PR #67, opened 2018-07-03
Use associative array when assigning new properties

mmachatschek:


mmachatschek:
@ihanli I hope this PR is still on your radar. Extending properties with a custom extension and correctly using `Property`-Class of libflexport would lead to a problem with the export file.

PR #74, opened 2018-07-11
Replace brackets from image urls

mmachatschek:
This replaces brackets in image urls which might appear on some shopware instances. URLs like `https://www.xyz.de/media/image/3a/83/55/88780_image1[1].jpg` breaks the XSD Schema requirements. This edgecase is not handled by the libflexport library at the moment. Issue was created [here](https://github.com/findologic/libflexport/issues/66). Still images can have non URL encoded links which will be fixed by this PR.

### DIFF
--- a/FinSearchAPI/BusinessLogic/Models/FindologicArticleModel.php
+++ b/FinSearchAPI/BusinessLogic/Models/FindologicArticleModel.php
@@ -335,8 +335,13 @@ class FindologicArticleModel
                 }
 
                 if (count($imageDetails) > 0) {
-                    $imagePath = $mediaService->getUrl($imageDefault);
-                    $imagePathThumb = $mediaService->getUrl(array_values($imageDetails)[0]);
+                    $replacements = array(
+                        '[' => '%5B',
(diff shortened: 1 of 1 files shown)


FinSearchAPI/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -335,8 +335,13 @@ protected function setImages()
                 }
 
                 if (count($imageDetails) > 0) {
-                    $imagePath = $mediaService->getUrl($imageDefault);
-                    $imagePathThumb = $mediaService->getUrl(array_values($imageDetails)[0]);
+                    $replacements = array(
```
ihanli: - [ ] Use `[]` notation instead of `array`. We can safely assume that the first one will be supported by the client's PHP
- [ ] Move this array to the beginning of this method. Since it stays the same for every image, there is no need to initialize it in every iteration.

mmachatschek: @ihanli Done! :+1:

PR #83, opened 2018-07-23
Disallow ajax search requests when findologic is active

mmachatschek:
@ihanli made this approach. Already works for one customer but we maybe should consider a better possibilty. Do you have an idea?

### DIFF
--- a/FinSearchUnified/Subscriber/Frontend.php
+++ b/FinSearchUnified/Subscriber/Frontend.php
@@ -5,6 +5,7 @@ namespace FinSearchUnified\Subscriber;
 use Enlight\Event\SubscriberInterface;
 use Enlight_Controller_ActionEventArgs;
 use Enlight_Event_EventArgs;
+use FinSearchUnified\Helper\StaticHelper;
 use FinSearchUnified\ShopwareProcess;
 
 class Frontend implements SubscriberInterface
@@ -36,7 +37,7 @@ class Frontend implements SubscriberInterface
         return [
             'Shopware_Controllers_Frontend_Search::indexAction::before' => 'beforeSearchIndexAction',
             'Enlight_Controller_Action_PreDispatch'                     => 'onPreDispatch',
-            'Enlight_Controller_Action_PreDispatch_Frontend_AjaxSearch' => 'onPreDispatchAjaxSearch',
+            'Enlight_Controller_Action_Frontend_AjaxSearch_Index' => 'onAjaxSearchIndexAction',
             'Enlight_Controller_Action_PostDispatchSecure_Frontend'     => 'onFrontendPostDispatch',
             'Enlight_Controller_Dispatcher_ControllerPath_Findologic'   => 'onFindologicController',
             'Enlight_Controller_Action_PreDispatch_Frontend_Listing'    => 'onFrontendListingPreDispatch',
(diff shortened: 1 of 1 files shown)


ihanli:
@mmachatschek @Weegy 
I'm not sure if this is the right approach, actually. The ajax search should be performed in any case but is supposed to use Shopware's original implementation of the product number search in case FINDOLOGIC is inactive. The latter should already be implemented. Is there an issue with that?

Weegy:
Normally it should trigger the productNumberSearch, so evrythibg should be fine

Outlook für iOS<https://aka.ms/o0ukef> beziehen
________________________________

mmachatschek:
@ihanli Yes seems like. The AjaxSearch Controller performs a standard search request with a different criteria object. As FINDOLOGIC is active at this time, the request is sent to FINDOLOGIC too. Can be reproduced with the current branch. I guess happens also in previous versions.

**Reproduce**

1. Add standard plugin
2. Activate plugin and add Shopkey
3. Go to fronend and type something in the searchfield
4. See network tab -> request is sent to the AjaxController
5. Product Search of Shopware is triggered
6. Request to FINDOLOGIC will be performed

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -36,6 +36,7 @@ public static function getSubscribedEvents()
         return [
             'Shopware_Controllers_Frontend_Search::indexAction::before' => 'beforeSearchIndexAction',
             'Enlight_Controller_Action_PreDispatch'                     => 'onPreDispatch',
+            'Enlight_Controller_Action_PreDispatch_Frontend_AjaxSearch' => 'onPreDispatchAjaxSearch',
```
ihanli:
```
'Shopware_Controllers_Frontend_AjaxSearch::indexAction::replace' => 'replaceAjaxSearchIndexAction'
```

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -136,4 +137,13 @@ public function onFindologicController(\Enlight_Event_EventArgs $args)
     {
         return $this->Path().'Controllers/Frontend/Findologic.php';
     }
+
+    public function onPreDispatchAjaxSearch()
```
ihanli:
```
    public function replaceAjaxSearchIndexAction(\Enlight_Hook_HookArgs $args)
    {
        if ((bool) Shopware()->Config()->get('ActivateFindologic')) {
            Shopware()->Container()->get('front')->Plugins()->ViewRenderer()->setNoRender();

            return true;
        }
    }
```

ihanli:
@mmachatschek 
Unfortunately I'm not able to make any code changes to your branch, so please check the review comments for a detailed implementation.

mmachatschek:
@ihanli 

following Error message appears when we do not set `->setnoRender()`

```
Ups! Ein Fehler ist aufgetreten!
Die nachfolgenden Hinweise sollten Ihnen weiterhelfen.

Unable to load template snippet 'frontend/ajax_search/index.tpl|frontend/fin_search_unified/header.tpl|frontend/plugins/seo/index.tpl' in custom/plugins/FroshProfiler/Components/Smarty/sysplugins/smarty_internal_templatebase.php on line 128
```

PR #88, opened 2018-07-25
Revert changes from 07f2613 and add additional check

mmachatschek:
@ihanli the changes from 07f2613 were not right. The method `setAllKeywords` expects an array of `Keyword` elements. The usergroup is set when creating a `Keyword`-object.

The `Attribute`-Class also expects an array of values (non-associative).


ihanli:
@mmachatschek Change the base branch to `develop`

PR #90, opened 2018-07-26
Last stock check on article

mmachatschek:
It looks like the `Detail` class has no `getLastStock` method. Calling the method via the `Article` class of the variant works.


ihanli:
@mmachatschek 
1. Remove the Zip file
2. Which version of Shopware was that? I just checked the class and it should indeed have that method.

mmachatschek:
@ihanli 
As I see the only the latest version of Shopware 5.4 supports the `getLastStock`-Method directly via the `Detail`-Class

ihanli:
@mmachatschek 
Seems like `Article::getLastStock` will be removed in Shopware 6, which will leave us with just `Detail::getLastStock`.

Let's do it this way:
1. Do a similar check in `ShopwareProcess::getAllProductsAsXmlArray` but only if the article has the method `::getLastStock` (use `method_exists`)
2. In `FindologicArticleModel::setVariantOrdernumbers` do sth. like
```
$lastStock = method_exists($detail, 'getLastStock') ? $detail->getLastStock() : $this->baseArticle->getLastStock();
```
This way we should be save in any case.

mmachatschek:
@ihanli updated condition

PR #93, opened 2018-08-01
Deactivate customization plugins of FINDOLOGIC

mmachatschek:


FinSearchUnified/FinSearchUnified.php
```diff
@@ -19,4 +22,45 @@ public function build(ContainerBuilder $container)
...
+            if ($plugin->getActive()) {
+                $pluginManager->deactivatePlugin($plugin);
+            }
+        } catch (\Exception $exception) {
+            Shopware()->PluginLogger()->info('ExtendFinSearchUnified plugin doesn\'t exist!');
+        }
+
+        try {
+            /** @var \Shopware\Models\Plugin\Plugin $plugin */
+            $plugin = $pluginManager->getPluginByName('FinSearchCustomized');
```
ihanli: This whole block may be removed since we agreed on leaving the name as is to avoid legacy plugins

mmachatschek: Done!

PR #94, opened 2018-08-01
Fetch only active products from database

mmachatschek:


FinSearchUnified/ShopwareProcess.php
```diff
@@ -131,7 +122,7 @@ public function getAllProductsAsXmlArray($selectedLanguage = 'de_DE', $start = 0
             }
 
             /** @var FindologicArticleFactory $findologicArticleFactory */
-            $findologicArticle = $findologicArticleFactory->create($article, $this->shopKey, $allUserGroups, $articleSales, $baseCategory);
+            $findologicArticle = $findologicArticleFactory->create($article, $this->shopKey, $allUserGroups, [], $baseCategory);
```
ihanli: Please remove the `$salesFrequency` argument since it has become obsolete. Therefore we also have to postpone this pull request for a later release until we know how to handle backward compatibility with `ExtendFinSearchUnified`

ihanli: @mmachatschek 
Ignore the last comment. We leave this as is.

mmachatschek:
@ihanli made one additional change

The salesfrequency expects the value to be an integer. The result of the DB query is a string so I casted the value to int again. Addtionally we could check for a positive integer, but negative sales isnt possible I guess so the need for that is not neccessary.

If no results are available, NULL is being returned and

```php
(int)NULL === 0
```

PR #101, opened 2018-08-04
Add productstream data to model

mmachatschek:
This PR depends on #100 

The product stream articles and the product stream categories are put into cache. This has to be done to increase performance while importing.

The cache will be cleared when `Article`, `Detail` and `ProductStream` Models are updated.

@ihanli **WIP**


mmachatschek:
@ihanli This PR needs extensive click testing. Because we have to make sure there are no memory leaks somewhere when caching a high amount of data like Product streams.

I also had the idea to not cache the categories because we can not split it up

ihanli:
Replaced by https://github.com/findologic/plugin-shopware-unified/pull/114

PR #128, opened 2018-09-14
Export rewrite

ihanli:


dblaichinger:
@Jeerrryyyy 
This is just a test!

dblaichinger:
@TheKeymaster This will probably never be continued, agree to close it?

TheKeymaster:
@dblaichinger as much as I would like to do this, I think that this will not be necessary for the Shopware 5 plugin, as the Shopware 6 plugin will eventually discontinue the development of the Shopware 5 plugin. At the same time there are currently more important topics, therefore I would not even have time to implement such an improvement.

*Closing* this PR for good.

PR #153, opened 2018-11-07
[FIXED] fixed remaining issues from previous pull request

zep-com:
In `PluginTest::testArticleExport` put the call of `PluginTest::createTestProduct` in a loop so we can get rid of the duplicate lines.
In `PluginTest::addDataProvider` use the case descriptions as keys for the parameter array. This will give you a more human readable message in case of a failing test. Also remove the case descriptions from the block comment. See the example below for the new data provider:
```
return [
    'both products are active and both are returned' => [[true, true], 2]
]
```
In `PluginTest::addDataProvider` provide a message for the assertion made in the actual test.
Did not use models to truncate data due to concurrency issues and indexes not being reset

### DIFF
--- a/FinSearchUnified/tests/PluginTest.php
+++ b/FinSearchUnified/tests/PluginTest.php
@@ -14,15 +14,6 @@ class PluginTest extends TestCase
         ],
     ];
 
-    /**
-     * @throws \Zend_Db_Adapter_Exception
-     */
-    protected function tearDown()
-    {
-        // Reset articles data after each test execution
-        $this->sResetArticles();
-    }
-
     public function testCanCreateInstance()
     {
         /** @var Plugin $plugin */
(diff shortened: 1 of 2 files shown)


FinSearchUnified/tests/PluginTest.php
```diff
@@ -114,7 +113,8 @@ private function createTestProduct($number, $isActive)
         $article = $resource->create($testArticle);
 
         // Assertion to make sure the test article is created properly
-        $this->assertInstanceOf(\Shopware\Models\Article\Article::class, $article);
+        $this->assertInstanceOf(\Shopware\Models\Article\Article::class, $article,
+            'FINDOLOGIC' . $number . ($isActive ? ": active " : " inactive ") . " article created");
```
ihanli: - [ ] Rephrase the message since it will be shown when the assertion fails.

- [ ] Make sure to use `sprintf` if you want to have variable values in the message.

zep-com: should I remove this assertion instead? as it generally throws an `Exception` if for any reason the article is not created

PR #156, opened 2018-11-14
Sw 210 shop search switch

zep-com:
Implement tests for shop search switch

### DIFF
--- a/FinSearchUnified/tests/Helper/StaticHelperTest.php
+++ b/FinSearchUnified/tests/Helper/StaticHelperTest.php
@@ -2,6 +2,7 @@
 
 namespace FinSearchUnified\tests\Helper;
 
+use FinSearchUnified\Constants;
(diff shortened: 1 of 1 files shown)


FinSearchUnified/tests/Helper/StaticHelperTest.php
```diff
@@ -0,0 +1,175 @@
...
+            ->getMock());
+
+        // Create Mock object for Shopware Config
+        Shopware()->Container()->set('config', $this->getMockBuilder('\Shopware_Components_Config')
+            ->setMethods(null)
+            ->disableOriginalConstructor()
+            ->getMock());
+
+        if ($isSearchPage !== null) {
+            $this->saveSession($sessionArray);
```
ihanli: `saveSession` and `saveConfig` shouldn't be necessary anymore since we are using mocks now. Instead, mock the respective methods and explicitly set the return values.

PR #165, opened 2018-11-27
Sw 216 fix sitemap cronjob

zep-com:


FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+        // Create mocked args for getting Subject and Request
+        $args = $this->getMockBuilder('\Enlight_Controller_ActionEventArgs')
+            ->setMethods(['getSubject', 'getRequest'])
+            ->getMock();
+        $args->method('getSubject')->willReturn($subject);
+        $args->method('getRequest')->willReturn($request);
+
+        $sessionArray = [
+            ['isSearchPage', !is_null($sSearch)],
```
ihanli: Try to use `isset` instead, it should have the same effect.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -0,0 +1,138 @@
...
+                'sModule' => 'backend'
+            ],
+            'Current Page is not Listing or Search' => [
+                'sSearch' => null,
+                'sCategory' => null,
+                'sController' => 'index',
+                'sAction' => 'index',
+                'sModule' => null
+            ],
+            'Check values after listingCount call on Search Page' => [
```
ihanli: I would suggest extracting these into a separate test.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -86,7 +94,18 @@ public static function configDataProvider()
                 'findologicDI' => false,
                 'isSearchPage' => false,
                 'isCategoryPage' => true,
+                'Request' => new RequestHttp(),
                 'expected' => false
+            ],
+            'FINDOLOGIC is active and Request is NULL' => [
```
ihanli: This should rather describe why the request is null e.g. `FINDOLOGIC is active but search was triggered via CLI`

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
```diff
@@ -0,0 +1,154 @@
...
+     * @param bool $isActiveForCategory
+     * @param bool $checkIntegration
+     * @param bool $isSearchPage
+     * @param bool $isCategoryPage
+     * @param bool $prepareCriteria
+     * @param int $expected
+     *
+     * @throws \Enlight_Exception
+     */
+    public function testCreateCriteria(
```
ihanli: Something looks odd here. It seems that you are testing `ProductStream\CriteriaFactory` before `ProductStream\Repository` is tested. That shouldn't be necessary since there is already a separate test class for `ProductStream\CriteriaFactory`. So:
* Remove everything that is not related to `ProductStream\Repository`.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -54,13 +53,17 @@ public function onFrontendPreDispatch(Enlight_Event_EventArgs $args)
...
+        $moduleName = $request->getModuleName();
+
         if ($this->isSearchPage($request)) {
             Shopware()->Session()->offsetSet('isSearchPage', true);
             Shopware()->Session()->offsetSet('isCategoryPage', false);
         } elseif ($this->isCategoryPage($request)) {
             Shopware()->Session()->offsetSet('isCategoryPage', true);
             Shopware()->Session()->offsetSet('isSearchPage', false);
-        } elseif ($this->isManufacturerPage($request)) {
+        } elseif ($this->isManufacturerPage($request) || $moduleName == 'backend' || ($controllerName !== 'listing'
```
ihanli: * Move everything `$controllerName` related to the next line, just to make it visually clear that they belong together.
* Use `===`. Always.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -696,8 +699,7 @@ public static function removeControlCharacters($value)
     public static function useShopSearch()
     {
         return (
-            !self::isFindologicActive() ||
-            self::checkDirectIntegration() ||
+            !self::isFindologicActive() || self::checkDirectIntegration() || Shopware()->Front()->Request() == null ||
```
ihanli: * Check for the request object right at the beginning. We actually don't care about the rest if there is no request anyway.
* Use `===`. Always.

zep-com: moving this in the beginning would cause error in the latter conditions as we are using `expects` in the mock object to use `atleastOnce`. We would have to remove that check or use `any` otherwise so it does not check the number of times the method is called

ihanli: @zep-com 
Unless you extract the case into a separate test?

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
zep-com: But it was mentioned in the ticket to implement test cases as for CriteriaFactory. What do you mean by that?

FinSearchUnified/Helper/StaticHelper.php
ihanli: @zep-com 
Yes, definitely but our experience shows that it is better to split such things up as it makes the tests more understandable and easier to maintain.
While your argument is definitely valid, you will have to trust me that it is better to have separate tests in the long run.
Also, since the method returns early for this case, you won't have to create as much mocks.

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+            ]
+        ];
+    }
+
+    /**
+     * @dataProvider shopSearchSwitchProvider
+     *
+     * @param bool $prepareCriteria
+     */
+    public function testIfPrepareCriteriaIsCalled($prepareCriteria)
```
ihanli: -> `testUsesOriginalOrDecoratedImplementation`

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $mockedRepository = $this->getMockBuilder('\Shopware\Components\ProductStream\Repository')
+            ->setMethods(['prepareCriteria'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        if ($prepareCriteria) {
+            $mockedRepository->expects($this->atLeastOnce())
```
ihanli: This should only be called once, so use `$this->once()` instead.

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+                ->method('prepareCriteria');
+        } else {
+            $mockedRepository->expects($this->never())
+                ->method('prepareCriteria');
+        }
+
+        $repository = new Repository($mockedRepository);
+        $repository->prepareCriteria(new Criteria(), 1);
```
ihanli: Why can the second parameter be  1? Shouldn't it cause an error since it is supposed to be of type `ShopContextInterface`?

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -247,8 +270,26 @@ public function testUseShopSearch(
...
+            ->getMock();
+        $front->expects($this->atLeastOnce())
+            ->method('Request')
+            ->willReturn(null);
+
+        // Assign mocked session variable to application container
+        Shopware()->Container()->set('front', $front);
+
+        $result = StaticHelper::useShopSearch();
+        $this->assertEquals(true, $result, 'Expected shop search to be triggered but findologic was triggered instead');
```
ihanli: Use `$this->assertTrue` instead.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -0,0 +1,182 @@
...
+     */
+    public function frontendPreDispatchProvider()
+    {
+        return [
+            'Search Page' => [
+                'sSearch' => 'Yes',
+                'sCategory' => null,
+                'sController' => 'index',
+                'sAction' => 'index',
+                'sModule' => null
```
ihanli: Use `frontend` instead of `null`

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * @dataProvider frontendPreDispatchProvider
+     *
+     * @param string $sSearch
+     * @param int|null $sCategory
+     * @param string $sController
+     * @param string $sAction
+     * @param string $sModule
+     */
+    public function testFrontendPreDispatchConditions($sSearch, $sCategory, $sController, $sAction, $sModule)
```
ihanli: I could be wrong but it seems that this method isn't actually testing anything. I can't see any assertions at least.

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
zep-com: No the second parameter is the `productStreamId` not the context. Unless I am mistaken, please let me know if I am creating the correct `Repository` object.

ihanli: Nope, you are correct. I just checked the wrong class.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
zep-com: ah yes, my bad. I think I forgot to add the assertions after we refactored a separate test method for the listingCount scenario.

PR #166, opened 2018-11-27
[BUGFIX] SW-215 trim category names

zep-com:


FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -188,6 +188,14 @@ public static function cleanStringProvider()
         ];
     }
 
+    public function categoryNamesProvider()
+    {
+        return [
+            'Root category name without children' => [1, ' Root '],
+            'Category name with parent' => [5, ' Genusswelten '],
```
ihanli: For this case, also wrap the parent category with spaces to make sure that all of them were trimmed

zep-com: @ihanli wrap the parent category as in? withe the `|` symbol?

PR #168, opened 2018-12-04
SW-213 build script for packaging repository

zep-com:


archive.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+SHOPWARE_DIRECTORY=../../../../shopware
```
ihanli: This script will never be executed inside of Shopware, so this will fail.

archive.sh
```diff
@@ -0,0 +1,20 @@
...
+
+VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plugin.xml)"
```
ihanli: Add some comments which describe what these two lines do.

archive.sh
```diff
@@ -0,0 +1,20 @@
...
+
+VERSION="$(echo -e "${VERSION_RAW}" | tr -d '[:space:]')"
+
+echo "Version: ${VERSION}"
+
+cp -rvf ./FinSearchUnified/ ${SHOPWARE_DIRECTORY}/tmp
```
ihanli: This has to be adapted accordingly since the shopware directory won't exist.

archive.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo "Copying files ... "
+
+cd FinSearchUnified
```
ihanli: I guess this is the wrong directory?

archive.sh
zep-com: then what would be the root directory?

archive.sh
zep-com: no, why? when you clone the repository you get the plugin directory inside it and since the archive.sh is in the root of the repository (project) I am doing a cd into the plugin directory to run the composer commands. Am I missing something here?

archive.sh
ihanli: the directory plugin-shopware-unified

archive.sh
ihanli: The original idea was:
* copy FinSearchUnified into `/tmp/FinSearchUnified`
* `cd` into `/tmp/FinSearchUnified`
* run the archive command

Right now you are copying the plugin into shopware's tmp folder but don't do anything with it.

PR #172, opened 2018-12-12
Reset categories active status after running tests

zep-com:


FinSearchUnified/Tests/BusinessLogic/ExportTest.php
```diff
@@ -49,6 +57,26 @@ protected function tearDown()
...
+     * @throws \Doctrine\ORM\OptimisticLockException
+     */
+    public static function tearDownAfterClass()
+    {
+        parent::tearDownAfterClass();
+
+        // Reset category active status which was used in the test cases
+        $categoryRepository = Shopware()->Models()->getRepository('Shopware\Models\Category\Category');
+
+        $categoryModels = $categoryRepository->findBy(['id' => [Shopware()->Shop()->getCategory()->getId(), 39]]);
```
ihanli: Use `$this->categoryRepository` instead instead.

zep-com: It is called inside a `static` method so $this wont be available

ihanli: You are absolutely right.

PR #173, opened 2019-01-10
Sw 217 implement smart did you mean

zep-com:


FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -83,4 +114,86 @@ private function setClientIp($ipAddress)
...
+            'revision',
+            $requestedUrl,
+            'revision was not found in the query parameters'
+        );
+        $this->assertArrayHasKey(
+            'shopkey',
+            $requestedUrl,
+            'shopkey was not found in the query parameters'
+        );
+        $this->assertEquals(
```
ihanli: Also check if the array key actually exists.

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -83,4 +114,86 @@ private function setClientIp($ipAddress)
...
+
+        $this->assertSame(
+            sprintf('https://service.findologic.com/ps/xml_2.0/%s/index.php', Shopware()->Shop()->getHost()),
+            $actualUrl,
+            'The resulting URL is not correct'
+        );
+        $this->assertArrayHasKey(
+            'userip',
+            $requestedUrl,
+            'userip was not found in the query parameters'
```
ihanli: Wrap `userip`, `revision` etc. with double quotes, so it is clear that this is the actualy name of the parameter.

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -83,4 +114,86 @@ private function setClientIp($ipAddress)
...
+        );
+        $this->assertArrayHasKey(
+            'shopkey',
+            $requestedUrl,
+            'shopkey was not found in the query parameters'
+        );
+        $this->assertEquals(
+            $forceOriginalQuery,
+            $requestedUrl['forceOriginalQuery'],
+            'forceOriginalQuery was not processed correctly '
```
ihanli: Remove the trailing white space.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -96,13 +123,29 @@ public function testFrontendPreDispatchConditions($sSearch, $sCategory, $sContro
...
+            ['findologicDI', false]
+        ]);
 
-        // Assign mocked session variable to application container
         Shopware()->Container()->set('session', $session);
 
+        $configArray = [
+            ['ActivateFindologic', true],
+            ['ActivateFindologicForCategoryPages', false],
+            ['findologicDI', false],
```
ihanli: `findologicDI`, `isSearchPage` and `isCategoryPage` only exist as session variables.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -430,5 +468,146 @@ private function updateParentCategoryName(Category $parent, Category $categoryMo
...
+     * @dataProvider finSmartDidYouMeanProvider
+     *
+     * @param string $didYouMeanQuery
+     * @param string $originalQuery
+     * @param string $queryStringType
+     * @param string $expectedType
+     * @param string $expectedAlternativeQuery
+     * @param string $expectedOriginalQuery
+     */
+    public function testFinSmartDidYouMean(
```
ihanli: Rename this to `testParsesSmartDidYouMeanData`. Otherwise it is hard to tell what this is actually testing.

FinSearchUnified/Tests/Functional/Controllers/Frontend/FinSearchUnified_Tests_Controllers_Frontend_SearchTest.php
```diff
@@ -0,0 +1,344 @@
...
+        Shopware()->Container()->set('fin_search_unified.product_number_search', $productNumberSearch);
+
+        $session = $this->getMockBuilder(Enlight_Components_Session_Namespace::class)
+            ->setMethods(['offsetGet', 'offsetExists'])
+            ->getMock();
+        $session->expects($this->atLeastOnce())->method('offsetExists')->willReturnMap([
+            'findologicDI' => true
+        ]);
+        $session->expects($this->atLeastOnce())->method('offsetGet')->willReturnMap([
+            ['isSearchPage', true],
```
ihanli: You could also set these directly instead of using a mock, since the values never change. Keep in mind that you will have to remove them again in `tearDown`.

FinSearchUnified/Tests/Functional/Controllers/Frontend/FinSearchUnified_Tests_Controllers_Frontend_SearchTest.php
```diff
@@ -0,0 +1,344 @@
...
+            ['isCategoryPage', false],
+            ['findologicDI', false]
+        ]);
+
+        Shopware()->Container()->set('session', $session);
+
+        $configArray = [
+            ['ActivateFindologic', true],
+            ['ActivateFindologicForCategoryPages', false],
+            ['findologicDI', false],
```
ihanli: `findologicDI`, `isSearchPage` and `isCategoryPage` are only available from the session.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -823,4 +823,30 @@ public static function setPromotion(SimpleXMLElement $xmlResponse)
             ]);
         }
     }
+
+    /**
+     * @param \SimpleXMLElement $xmlResponse
```
ihanli: If the namespace is already imported, make sure to remove the preceding backslash.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -823,4 +823,30 @@ public static function setPromotion(SimpleXMLElement $xmlResponse)
...
+     */
+    public static function setSmartDidYouMean(SimpleXMLElement $xmlResponse)
+    {
+        $query = $xmlResponse->query;
+        $originalQuery = (string)$query->originalQuery;
+        $didYouMeanQuery = (string)$query->didYouMeanQuery;
+        $queryString = $query->queryString;
+        $queryStringType = (string)$queryString->attributes()->type;
+
+        if (((isset($originalQuery) && $originalQuery !== '') || (isset($didYouMeanQuery) && $didYouMeanQuery !== ''))
```
ihanli: Instead of `(isset($originalQuery) && $originalQuery !== '')` use `!empty($originalQuery)`. Should yield the same result.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -823,4 +823,30 @@ public static function setPromotion(SimpleXMLElement $xmlResponse)
...
+            && $queryStringType !== 'forced') {
+            /** @var \Enlight_View_Default $view */
+            $view = Shopware()->Front()->Plugins()->ViewRenderer()->Action()->View();
+            $type = isset($didYouMeanQuery) && !empty($didYouMeanQuery) ? 'did-you-mean' : $queryStringType;
```
ihanli: The call for `isset` shouldn't be necessary as `empty` already covers what you are trying to check.

FinSearchUnified/Tests/Components/ProductStream/RepositoryTest.php
```diff
@@ -47,10 +50,64 @@ public function shopSearchSwitchProvider()
...
+        $shopKey,
+        $isActiveForCategory,
+        $checkIntegration,
+        $isSearchPage,
+        $isCategoryPage,
+        $prepareCriteria
+    ) {
+        $request = new \Enlight_Controller_Request_RequestHttp();
+        if ($isCategoryPage) {
+            $request->setParam('sCategory', 3);
```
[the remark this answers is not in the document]
zep-com: @ihanli in one of the other stories you mentioned that in order for the search to know its a category page, it checks for the sCategory parameter in the request. So that is why I kept it in this condition if its a category page so it renders the correct behaviour

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -430,5 +468,146 @@ private function updateParentCategoryName(Category $parent, Category $categoryMo
...
+            ->setMethods(['Plugins'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $front->method('Plugins')
+            ->willReturn($plugin);
+
+        Shopware()->Container()->set('front', $front);
+
+        // Get assigned variable from mocked view object
+        $finSmartDidYouMean = Shopware()
```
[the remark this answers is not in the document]
zep-com: @ihanli If I am mocking the `View` itself, wont it be necessary to mock the `getAssign` method as well? Because otherwise it will simply return null if we haven't mocked the method. And if we are mocking it then we would need to define a return value as well, which in turn would be testing our mocked data indirectly.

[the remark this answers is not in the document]
ihanli: @zep-com Not quite. You can test the passed parameter of your mock's `assign` method. So you won't have to mock `getAssign` + `StaticHelper` would use the mocked view object e.g.
```
$view->expects($this->once())
                ->method('assign')
                ->with($this->callback(function ($data) {
                   \PHPUnit_Framework_Assert::assertEquals(
                        'some value',
                        $data['some-key']
                    );
                   
                    return true;
                }));
```

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -83,4 +101,99 @@ private function setClientIp($ipAddress)
...
+            ->setMethods(['offsetGet'])
+            ->getMock();
+        $session->expects($this->atLeastOnce())
+            ->method('offsetGet')
+            ->willReturnMap($sessionArray);
+
+        // Assign mocked session variable to application container
+        Shopware()->Container()->set('session', $session);
+
+        $_GET[Constants::SDYM_PARAM_FORCE_QUERY] = $forceOriginalQuery;
```
ihanli: Use the string instead of the constant here. Imagine that we change its value but accidentally add a typo while doing so. We would never notice that we are sending the wrong parameter since we are using the constant everywhere.

So, constants are good. Use them in the code where applicable but still consider using strings in tests instead of constants. This way we will notice soon enough if the parameter name was changed to something wrong.

Also change other occurrences of `Constants::SDYM_PARAM_FORCE_QUERY` in this test.

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
ihanli: The first parameter of `assertArrayHasKey` can be in single quotes. What I meant was that the assertion message should look something like `"userip" was not found in the query parameters`.

FinSearchUnified/Resources/views/frontend/search/fuzzy.tpl
```diff
@@ -0,0 +1,30 @@
...
+                    "?
+                {/s}
+            {elseif $finSmartDidYouMean.type == 'corrected'}
+                {s name='frontend/search/fuzzy/corrected_query'}
+                    No results for "{$finSmartDidYouMean.original_query}". Showing results for "{$finSmartDidYouMean.alternative_query}" instead
+                {/s}
+        {/if}
+        </p>
+    {else}
+        <p></p>
```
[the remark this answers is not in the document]
zep-com: @ihanli actually you said you wanted an empty `<p>` tag for styling purposes so it would be possible to render that in the `{else} of the outer `{if}` as we already have the `fl-smart-did-you-mean` id in the inner `{if}` block

PR #176, opened 2019-01-28
Add filter for reverse proxy header

mmachatschek:
Here's a wikipedia link to the implemented changes: https://en.wikipedia.org/wiki/X-Forwarded-For


ihanli:
@mmachatschek 
Will be postponed until planned in detail. This PR will be closed but used for future reference.

dblaichinger:
@mmachatschek 
Isi has this change on his list and will prepare a story for it for the next sprint.
The change will not get lost in the depths of github.

PR #177, opened 2019-02-04
SW-228 get client IP from proxy

zep-com:


FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -77,10 +105,50 @@ public function testSendOnlyUniqueUserIps($unfilteredIp, $expectedValue)
     }
 
     /**
+     * @dataProvider reverseProxyIpAddressProvider
+     *
+     * @param string $unfilteredIp
```
ihanli: Shouldn't this be of type `array`?

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
```diff
@@ -77,10 +105,50 @@ public function testSendOnlyUniqueUserIps($unfilteredIp, $expectedValue)
...
+
+    public function testHandlesUnknownClientIp()
+    {
+        $urlBuilder = new UrlBuilder($this->httpClient);
+
+        $criteria = new Criteria();
+        $criteria->offset(0)->limit(2);
+
+        $urlBuilder->buildQueryUrlAndGetResponse($criteria);
+        /** @var Zend_Uri_Http $requestedUrl */
```
ihanli: This doesn't seem correct anymore. I guess this could be removed since `$requestedUrl` is just an array?

FinSearchUnified/Tests/Helper/UrlBuilderTest.php
zep-com: @ihanli No its a comma separated string actually. I did get confused earlier that why do you want me to type hint the `string`, but I thought if you want for this case then its fine with me.

PR #180, opened 2019-02-08
Sw 234 use findologic querybuilder factory

zep-com:


FinSearchUnified/Tests/Helper/Utility.php
```diff
@@ -62,4 +62,17 @@ public static function sResetArticles()
...
         }
     }
+
+    /**
+     * Allows to set a Shopware config
+     *
+     * @param string $name
+     * @param mixed $value
+     */
+    public static function setConfig($name, $value)
```
ihanli: Create class `FinSearchUnified\Tests\TestBase` which extends `TestCase` and move `setConfig` there (also make it protected instead of public).
Each unit test that uses `setConfig` should extend `TestBase` (ignore controller tests).

PR #189, opened 2019-02-28
Sw 245 correctly decode smart suggest parameter

zep-com:


FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -74,6 +68,16 @@ public function listingCountConditionProvider()
         ];
     }
 
+    public function vendorProvider()
+    {
+        return [
+            'setRequestUri contains vendor=Brand' => ['Brand'],
```
ihanli: Descriptions of data sets shouldn't be too technical. Change these to:
* `Regular vendor name`
* `Vendor name containing whitespace`
* `Vendor name containing "+" character`
* `Vendor name containing special characters`

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -157,54 +156,86 @@ public function testSessionValuesAfterListingCount($sSearch, $sCategory, $sContr
...
 
+        $frontend = Shopware()->Container()->get('fin_search_unified.subscriber.frontend');
         $frontend->onFrontendPreDispatch($args);
 
+        // Check session values after FrontendPreDispatch Call
+        $isCategoryPage = Shopware()->Session()->isCategoryPage;
+        $isSearchPage = Shopware()->Session()->isSearchPage;
+
         $this->assertEquals(
+            $isCategory,
```
ihanli: Check the argument list of method `assertEquals` and make sure that `$isCategory` and `$isCategoryPage` are in the correct order.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -157,54 +156,86 @@ public function testSessionValuesAfterListingCount($sSearch, $sCategory, $sContr
...
+     * @dataProvider vendorProvider
+     *
+     * @param string $vendor
+     */
+    public function testBeforeSearchIndexAction($vendor)
+    {
+        Shopware()->Session()->isCategoryPage = null;
+        Shopware()->Session()->isSearchPage = true;
+
+        $attrib['vendor'] = $vendor;
```
ihanli: * This will work since PHP is intuitive enough to know that `$attrib` should be an array but you should still explicitly declare it.
* The value of `$attrib` is not quite correct.

Use this instead:
```
$attrib = [
    'vendor` => [$vendor]
];
```

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -157,54 +156,86 @@ public function testSessionValuesAfterListingCount($sSearch, $sCategory, $sContr
...
+
+        // Create Request object to be passed in the mocked Subject
+        $request = new RequestHttp();
+        $request->setControllerName('index')
```
ihanli: Change the controller's name to `search`. This won't affect the outcome of the test but it reflects the use case better.

PR #191, opened 2019-03-07
SW-252 Gracefully handle vendors without names during export

mmachatschek:


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -527,9 +527,12 @@ protected function setAttributes()
         /** @var Product\Manufacturer $supplier */
         $supplier = $this->productStruct->getManufacturer();
         if ($supplier) {
-            $xmlSupplier = new Attribute('brand');
-            $xmlSupplier->setValues([StaticHelper::removeControlCharacters($supplier->getName())]);
-            $allAttributes[] = $xmlSupplier;
+            $supplierName = trim(StaticHelper::removeControlCharacters($supplier->getName()));
```
ihanli: Check if `\FinSearchUnified\Helper\StaticHelper::cleanString` could be used instead (since it calls `trim` and `\FinSearchUnified\Helper\StaticHelper::removeControlCharacters` anyway).

mmachatschek: @ihanli i guess we can call `\FinSearchUnified\Helper\StaticHelper::cleanString` as it removes potential bad characters as well.

mmachatschek:
@ihanli updated the PR

> Seems that the tests are missing.

I couldn't find any tests for the export except for the Statichelper I'm using which is already tested. How should I proceed?

ihanli:
@mmachatschek 
* Check `\FinSearchUnified\Tests\PluginTest::createTestProduct` to see how to create products.
* Create test class `FinSearchUnified\Tests\BusinessLogic\Models\FindologicArticleModelTest`.
  * Create an instance of `\FinSearchUnified\BusinessLogic\Models\FindologicArticleModel`.
  * Provide any data necessary so you can successfully call `\FinSearchUnified\BusinessLogic\Models\FindologicArticleModel::getXmlRepresentation`.
  * Use the return value for assertions.
* Since you are only interested in the `brand` field, your setup can be as simple as possible.

mmachatschek:
@ihanli Ready for another round of review. I had to add the supplier with a space as a name, as creating a supplier with a name of `NULL` isn't possible with the current setup.

I copied the `Utility::class` from the `Tests\Helper` folder of the PR #173 as discussed.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->manager = new Manager();
+        $this->articleFactory = new FindologicArticleFactory();
+    }
+
+    /**
+     * Method to create test products for the export
+     *
+     * @param array $testProductConfiguration
+     *
+     * @return \Shopware\Models\Article\Article|null
```
ihanli: Import class `Shopware\Models\Article\Article` and rewrite this to `@return Article|null`.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private function createTestProduct($testProductConfiguration)
```
ihanli: Add the type hint here.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+            ]
+        ];
+    }
+
+    /**
+     * @dataProvider articleSupplierProvider
+     *
+     * @param array $articleConfiguration The article configuration with the corresponding supplier.
+     */
+    public function testEmptySuppliersAreSkipped($articleConfiguration)
```
ihanli: Add the type hint here.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+                [],
+                $baseCategory
+            );
+            $this->assertEquals(get_class($findologicArticle), FindologicArticleModel::class);
+        } catch (\Exception $exception) {
+            $this->fail('Exception may not be thrown here!');
+        }
+    }
+
+    protected function tearDown()
```
ihanli: Move this near the setup method.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+            $findologicArticle = $this->articleFactory->create(
+                $articleFromConfiguration,
+                'ABCD0815',
+                [],
+                [],
+                $baseCategory
+            );
+            $this->assertEquals(get_class($findologicArticle), FindologicArticleModel::class);
+        } catch (\Exception $exception) {
+            $this->fail('Exception may not be thrown here!');
```
ihanli: There are two issues with this try/catch:
* `TestCase::assertEquals` throws an exception when it fails. So you will always get `Exception may not be thrown here!`, even if `$this->articleFactory->create` didn't actually throw one.
* The try/catch logic can be removed completely since an unexpected exception causes the test to fail anyway.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+
+    protected function tearDown()
+    {
+        Utility::sResetArticles();
```
ihanli: Also call the parent method.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * @dataProvider articleSupplierProvider
+     *
+     * @param array $articleConfiguration The article configuration with the corresponding supplier.
+     */
+    public function testEmptySuppliersAreSkipped($articleConfiguration)
+    {
+        try {
+            $baseCategory = new Category();
```
ihanli: Check if this also creates a DB entry. If yes -> consider removing it in the tearDown method.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: I checked this. As the category is only created temporarily, its not being persistet. I debugged it and checked the database to make sure also.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: Done!

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
ihanli: Ehm, I don't see a change?

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
ihanli: Same, here. I don't see a change.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: @ihanli i missunderstood this, thought of the type hints in the docblock. Should be done now.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
mmachatschek: @ihanli i missunderstood this, thought of the typehints in the docblock.

PR #193, opened 2019-03-08
SW-248-fix-sitemap-generation-backend

zep-com:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -13,30 +14,27 @@
...
     /**
      * Creates a product search result for the passed criteria object.
      * The criteria object contains different core conditions and plugin conditions.
      * This conditions has to be handled over the different condition handlers.
-     *
      * The search gateway has to implement an event which plugin can be listened to,
      * to add their own handler classes.
      *
-     * @param \Shopware\Bundle\SearchBundle\Criteria                        $criteria
+     * @param \Shopware\Bundle\SearchBundle\Criteria $criteria
```
ihanli: This doesn't need to be the complete class path since `Criteria` has been imported use `@param Criteria $criteria` instead

PR #197, opened 2019-03-18
SW-250-missing-methods-in-querybuilder

zep-com:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderTest.php
```diff
@@ -185,4 +185,169 @@ public function testQuerybuilderRequestUrl($isSearch, $endpoint)
...
+
+    /**
+     * @dataProvider vendorFilterProvider
+     *
+     * @param array $filters
+     * @param array $expectedFilters
+     *
+     * @throws Exception
+     */
+    public function testAddFilterMethod($filters, $expectedFilters)
```
ihanli: Add the missing typehints to the argument list.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
```diff
@@ -166,4 +166,105 @@ private function getClientIp()
...
+    }
+
+    /**
+     * @param string $key
+     *
+     * @return mixed|null
+     */
+    private function getParameter($key)
+    {
+        if (!isset($this->parameters['attrib']) || !isset($this->parameters['attrib'][$key])) {
```
ihanli: Since the string `attrib` is used quite often:
* Extract it into a constant (can be a member of QueryBuilder), e.g. `QueryBuilder::PARAMETER_KEY_ATTRIB`.
* Make sure to use the constant only here, not in the tests.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
```diff
@@ -166,4 +166,105 @@ private function getClientIp()
...
+     * @param mixed $value
+     */
+    private function addParameter($key, $value)
+    {
+        if (!isset($this->parameters['attrib'])) {
+            $this->parameters['attrib'] = [];
+        }
+
+        if (!isset($this->parameters['attrib'][$key])) {
+            $this->parameters['attrib'][$key] = $value;
```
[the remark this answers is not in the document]
zep-com: Depends on your use case actually. For e.g the parameters are already being passed as an array, such as categories. 
However price is being passed directly as an associated array. 
`$this->addParameter('price', ['min' => urldecode($min), 'max' => urldecode($max)]);`

But then filters' have the first entry as an array
```
if (!empty($this->getParameter($key))) {
   $this->addParameter($key, urldecode($value));
} else {
   $this->addParameter($key, [urldecode($value)]);
}
```

so all these different parameters have different use cases.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
```diff
@@ -166,4 +167,102 @@ private function getClientIp()
...
+     * @param mixed $value
+     */
+    private function addParameter($key, $value)
+    {
+        if (!isset($this->parameters[self::PARAMETER_KEY_ATTRIB])) {
+            $this->parameters[self::PARAMETER_KEY_ATTRIB] = [];
+        }
+
+        if (!isset($this->parameters[self::PARAMETER_KEY_ATTRIB][$key])) {
+            $this->parameters[self::PARAMETER_KEY_ATTRIB][$key] = [$value];
```
ihanli: Unfortunately you changed this too early. I was about to say that you are right and this doesn't need to be changed.

zep-com: No problem, will revert it again and the test as well.

PR #199, opened 2019-03-19
Sw 251 sorting handlers productnumbersearch

zep-com:


FinSearchUnified/Bundle/SearchBundleFindologic/FindologicSortingHandlerInterface.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace FinSearchUnified\Bundle\SearchBundleFindologic;
+
+use Shopware\Bundle\SearchBundle\SortingInterface;
+use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
+
+interface FindologicSortingHandlerInterface
```
ihanli: Rename this interface to `SortingHandlerInterface`. Its namespace is already mentioning the relation to FINDOLOGIC multiple times.

FinSearchUnified/Bundle/SearchBundleFindologic/SortingHandler/PopularitySortingHandler.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Handles the passed sorting object.
+     *
+     * @param SortingInterface $sorting
+     * @param QueryBuilder $query
+     * @param ShopContextInterface $context
+     */
+    public function generateSorting(SortingInterface $sorting, QueryBuilder $query, ShopContextInterface $context)
+    {
+        /** @var PopularitySorting $sorting */
```
ihanli: Why is this necessary? Shouldn't the IDE already know about the accessible methods (since it implements the interface)?

Please have a second look and remove the line from each sorting handler if it isn't needed.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/SortingHandler/PopularitySortingHandlerTest.php
```diff
@@ -0,0 +1,67 @@
...
+use Shopware\Components\Test\Plugin\TestCase;
+
+class PopularitySortingHandlerTest extends TestCase
+{
+    /**
+     * @var QueryBuilder
+     */
+    private $querybuilder;
+
+    private $context;
```
ihanli: Add the missing PHPDoc (applies for each test class).

FinSearchUnified/Bundle/SearchBundleFindologic/SortingHandler/PopularitySortingHandler.php
zep-com: Please check the comment for the same in condition handler PR

ihanli: Ah, okay. It would work without it if we used the abstract sorting class instead of the interface. Anyhow, it can stay the way it is.

PR #202, opened 2019-04-01
SW-256-export-product-stream-for-categories-with-children

zep-com:


FinSearchUnified/Tests/ShopwareProcessTest.php
```diff
@@ -0,0 +1,110 @@
...
+        );
+
+        $reflector = new ReflectionObject($shopwareProcess);
+        $method = $reflector->getMethod('parseProductStreams');
+        $method->setAccessible(true);
+        $articles = $method->invoke($shopwareProcess, $parameters);
+
+        foreach ($articles as $article) {
+            foreach ($article as $category) {
+                $this->assertThat($category, $this->logicalOr(
```
ihanli: * Do not only check if the categories are present but also if the correct categories are assigned to the product (by checking the array index which should be the product's ID).
* Theoretically the `logicalOr` shouldn't be necessary since `$categoryB` and `$categoryD` should always be present.

So for the assertions:
* Check if the keys of `$articles` contain only the IDs of the created base products.
* Check if each entry of `$articles` has both `$categoryB` and `$categoryD`.

FinSearchUnified/Tests/ShopwareProcessTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+        $mockSearchResult = $this->createMock(ProductNumberSearchResult::class);
+        $mockSearchResult->method('getProducts')->willReturn($result->products);
+
+        $contextService = Shopware()->Container()->get('shopware_storefront.context_service');
+
+        $mockProductNumberSearch = $this->createMock(ProductNumberSearch::class);
+        $mockProductNumberSearch->expects($this->atLeastOnce())->method('search')->willReturn($mockSearchResult);
+
+        /** @var Repository $mockRepository */
```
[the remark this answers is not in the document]
zep-com: If I move it there then the mock method gives the warning about method "method" not found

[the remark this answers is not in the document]
ihanli: That's because class `Repository` doesn't have a method called `method`. We have two options here:
* Either move this to where the variable is initialized and use the mock class instead of `Repository`
* Or remove this entirely since `$mockRepository` isn't explicitly used from here on.

Both work for me, I will leave the decision to you.

FinSearchUnified/Tests/ShopwareProcessTest.php
```diff
@@ -0,0 +1,116 @@
...
+            $mockProductNumberSearch
+        );
+
+        $reflector = new ReflectionObject($shopwareProcess);
+        $method = $reflector->getMethod('parseProductStreams');
+        $method->setAccessible(true);
+        $articles = $method->invoke($shopwareProcess, $parameters);
+
+        $this->assertEmpty(
+            array_diff_key(array_column($products, 'id'), $articles),
```
ihanli: Nice implementation.

FinSearchUnified/Tests/ShopwareProcessTest.php
```diff
@@ -0,0 +1,116 @@
...
+        foreach ($articles as $articleID => $categories) {
+            $this->assertSame(
+                $categoryD,
+                $categories[0],
+                'Expected categoryD to be assigned to the article'
+            );
+            $this->assertSame(
+                $categoryB,
+                $categories[1],
+                'Expected categoryB to be the parent of the assigned categoryD'
```
ihanli: Doesn't accurately reflect what we are testing. Maybe `Expected parent of categoryD to be assigned to the article`?

PR #203, opened 2019-04-04
SW-255-fix-live-reloading

zep-com:


FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function beforeListingCountAction(\Enlight_Hook_HookArgs $args)
+    {
+        /** @var \Shopware_Controllers_Widgets_Listing $subject */
+        $subject = $args->getSubject();
+
+        $params = $subject->Request()->getParams();
+
+        if (!isset($params['sSearch']) && !isset($params['sCategory'])) {
```
ihanli: I didn't consider this before but `\FinSearchUnified\Helper\StaticHelper::useShopSearch` should also be checked. So at the beginning of the method, if `useShopSearch` is `true` you can return without any further computation.

Adapt the test `testBeforeListingCountAction` so that `useShopSearch` returns false and implement a separate test which checks the behavior for:
* `useShopSearch` -> `false`
* `sSearch` isn't present
* `sCategory` isn't present

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -0,0 +1,32 @@
...
+     * @return array The event names to listen to
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            'Shopware_Controllers_Widgets_Listing::listingCountAction::before' => 'beforeListingCountAction'
+        ];
+    }
+
+    public function beforeListingCountAction(\Enlight_Hook_HookArgs $args)
```
ihanli: Import `Enlight_Hook_HookArgs` so you can omit the leading backslash.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -0,0 +1,96 @@
...
+use Shopware_Controllers_Widgets_Listing;
+
+class WidgetsTest extends TestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        // reset global variable as it conflicts with existing tests in this class
+        $_GET = [];
```
ihanli: This should be done in the test class which actually sets `$_GET`.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $subject = Shopware_Controllers_Widgets_Listing::Instance(
+            Shopware_Controllers_Widgets_Listing::class,
+            [
+                $request,
+                new Enlight_Controller_Response_ResponseHttp()
+            ]
+        );
+
+        // Create mocked args for getting Subject and Request
+        $args = $this->getMockBuilder(Enlight_Hook_HookArgs::class)
```
ihanli: It should be possible to use the class directly instead of a mock.

FinSearchUnified/Subscriber/Widgets.php
zep-com: @ihanli how will this be implemented? if findologic search is active, then it will always have sSearch present if neither category or search is provided. Or did you mean useShopSearch -> true

ihanli: Yes, you are right. I updated the previous comment.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
zep-com: I remember why I mocked it, due to backward compatibility as the test were failing for older shopware versions

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -4,14 +4,20 @@
...
 use Shopware\Components\Test\Plugin\TestCase;
 use Shopware_Controllers_Widgets_Listing;
-use Enlight_Controller_Response_ResponseHttp;
 
 class FrontendTest extends TestCase
 {
+    public static function tearDownAfterClass()
+    {
+        parent::tearDownAfterClass();
+        $_GET = [];
```
ihanli: Move this to the `tearDown` method, just in case it would affect multiple tests.

PR #205, opened 2019-04-04
Sw 239 querybuilder factory productnumbersearch

zep-com:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+     * @throws Exception
+     */
+    public function testCreateQueryWithoutConditions()
+    {
+        $criteria = new Criteria();
+
+        $query = $this->factory->createQuery($criteria, $this->context);
+        $params = $query->getParameters();
+
+        $this->assertArrayHasKey('group', $params, 'Usergroup was expected to be present in the parameters');
```
ihanli: Add an assertion which checks the value of `$params['group']`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+        $this->assertArrayNotHasKey('attrib', $params, 'No attributes were expected to be present in the parameters');
+        $this->assertArrayNotHasKey('query', $params, 'No search query was expected to be present in the parameters ');
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function testCreateQueryithConditions()
```
ihanli: There seems to be a typo in the method name.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+        $this->assertArrayHasKey('cat', $attrib, 'Category was expected to be present in the attribute parameters');
+        $this->assertEquals(
+            ['Genusswelten', 'Genusswelten_Tees und Zubehör_Tees'],
+            $attrib['cat'],
+            'Expected categories to contain the name of the provided category IDs'
+        );
+
+        // PriceCondition
+        $this->assertArrayHasKey('price', $attrib, 'Prices were expected to be present in the attribute parameters');
+        $this->assertEquals(1, $attrib['price']['min'], 'Expected minimum price to be 1');
```
ihanli: Add assertions which check if the keys `min` and `max` are actually present.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+        $this->assertEquals(20, $attrib['price']['max'], 'Expected maximum price to be 20');
+
+        // ProductAttributeCondition
+        $this->assertArrayHasKey('vendor', $attrib, 'Expected "vendor" to be available in the attribute parameters');
+        $this->assertEquals(
+            ['Findologic Rockers'],
+            $attrib['vendor'],
+            'Expected vendor to be an array containing "Findologic Rockers"'
```
ihanli: It would say that this should either be changed to `"vendor"` or you change it to `vendor` in the previous assertion. Both options work for me but it should be consistent.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @dataProvider sortingProvider
+     *
+     * @param SortingInterface $sorting
+     * @param string $expected
+     *
+     * @throws Exception
+     */
+    public function testCreateQueryWithSingleSorting($sorting, $expected)
```
ihanli: Add the typehint to `$sorting`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+        $criteria->addSorting(new PopularitySorting());
+        $criteria->addSorting(new SimpleSorting('name'));
+        $criteria->addCondition(new SearchTermCondition('blubbergurke'));
+        $criteria->addCondition(new IsAvailableCondition());
+
+        $query = $this->factory->createQueryWithSorting($criteria, $this->context);
+        $params = $query->getParameters();
+
+        $this->assertArrayHasKey('order', $params, 'Sorting was expected to be present in the parameters');
+        $this->assertArrayHasKey('query', $params, 'Search query was expected to be present in the parameters ');
```
ihanli: Add an assertion which checks the value.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,180 @@
...
+     */
+    public function testCreateQueryWithSingleSorting($sorting, $expected)
+    {
+        $criteria = new Criteria();
+        $criteria->addSorting($sorting);
+
+        $query = $this->factory->createQueryWithSorting($criteria, $this->context);
+        $params = $query->getParameters();
+
+        $this->assertSame($expected, $params['order'], sprintf('Expected sorting to be "%s"', $expected));
```
ihanli: This passes because of the way PHP works but doesn't accurately represent what we are trying to test. I would suggest that:
* If `$expected` is `null`:
  * Check that key `order` doesn't exist in `$params`.
* If `$expected` is not `null`:
  * Check for the value of `$params['order'] (already done).
  * Check that key `order` exists in `$params`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -0,0 +1,282 @@
...
+        $this->assertSame('salesfrequency ASC', $params['order'], 'Expected sorting order to be "salesfrequency ASC"');
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function testCreateProductQueryWithOffsetAndLimit()
+    {
+        $criteria = new Criteria();
+        $criteria->offset(5);
```
ihanli: What about the case where offset is 0 and limit 1?

zep-com: you said to remove those test cases as we have setup the condition to make the limit 0 on this condition

ihanli: I said that you can drop the case where limit was supposed to be zero since the min value for it is 1. `offset = 0` and `limit = 1` is still a valid test case.

zep-com: I see, okay

PR #209, opened 2019-04-12
Separate querybuilder for navigation and search

zep-com:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private $installerService;
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->config = Shopware()->Container()->get('config');
+        $this->installerService = Shopware()->Container()->get('shopware_plugininstaller.plugin_manager');
+        Shopware()->Session()->offsetSet('isSearchPage', false);
```
ihanli: Implement a tearDown method (should be placed right after setUp) which unsets `isSearchPage`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $shopUrl = rtrim(Shopware()->Shop()->getHost(), '/');
+        $parameters = [
+            'outputAdapter' => 'XML_2.0',
+            'userip' => 'UNKNOWN',
+            'revision' => $plugin->getVersion(),
+            'shopkey' => $this->config->offsetGet('ShopKey')
+        ];
+        $aliveUrl = sprintf(
+            '%s/%s/%s?shopkey=%s',
+            QueryBuilder::BASE_URL,
```
ihanli: I would prefer that you use the actual value instead of the constant. Which also means that you don't have to provide it as parameter but can directly put it into the string.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $shopUrl,
+            QueryBuilder::ALIVE_ENDPOINT,
```
ihanli: I would prefer that you use the actual value instead of the constant. Which also means that you don't have to provide it as parameter but can directly put it into the string.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $parameters['shopkey']
+        );
+        $executeUrl = sprintf(
+            '%s/%s/%s?%s',
+            QueryBuilder::BASE_URL,
```
ihanli: I would prefer that you use the actual value instead of the constant. Which also means that you don't have to provide it as parameter but can directly put it into the string.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $shopUrl,
+            NavigationQueryBuilder::ENDPOINT,
```
ihanli: I would prefer that you use the actual value instead of the constant. Which also means that you don't have to provide it as parameter but can directly put it into the string.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
```diff
@@ -0,0 +1,112 @@
...
+            http_build_query($parameters)
+        );
+        $httpResponse = new Response(200, [], 'alive');
+        $httpClientMock = $this->createMock(GuzzleHttpClient::class);
+        $aliveCallback = $this->returnCallback(function ($url) use ($aliveUrl, $httpResponse) {
+            Assert::assertSame($aliveUrl, $url);
```
ihanli: Since we are checking the provided parameter, this assertion should be made in the `with` part. So check if something like this is possible:
```
$httpClientMock->expects($this->exactly(2))
    ->method('get')
    ->with($this->onConsecutive([THE CALLBACKS GOE HERE]))
    ->will($this->onConsecutiveCalls($httpResponse, $httpResponse));
```
See  `\FinSearchUnified\Tests\Subscriber\FrontendTest::testBeforeSearchIndexAction` for reference.

If it turns out that this approach won't work, we can stick with the current implementation.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -278,6 +290,22 @@ public function testCreateProductQueryWithOffsetAndLimit($offset, $expectedOffse
...
+     * @param string $expectedInstance
+     *
+     * @throws Exception
+     */
+    public function testTypeOfQueryBuilder($isSearchPage, $expectedInstance)
+    {
+        Shopware()->Session()->offsetSet('isSearchPage', $isSearchPage);
+
+        $builder = $this->factory->createQueryBuilder();
+        $this->assertInstanceOf($expectedInstance, $builder, sprintf('Expected %s to be returned', $expectedInstance));
```
ihanli: The third parameter won't be necessary since `assertInstanceOf` will already give you a similar message.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilderFactoryTest.php
```diff
@@ -302,4 +330,12 @@ public function offsetLimitProvider()
             'Offset is 5 and Limit = 2' => [5, 5, 2, 2],
         ];
     }
+
+    public function isSearchPageDataProvider()
+    {
+        return [
+            'isSearchPage is true' => [true, SearchQueryBuilder::class],
```
ihanli: Rename the data sets to `Search request` and `Navigation request`.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -11,12 +11,12 @@
 
 class WidgetsTest extends TestCase
 {
-    protected static $ensureLoadedPlugins = [
-        'FinSearchUnified' => [
-            'ShopKey' => '0000000000000000ZZZZZZZZZZZZZZZZ',
-            'ActivateFindologic' => true
-        ],
-    ];
+    public static function setUpBeforeClass()
```
ihanli: Why was this change necessary? `$ensureLoadedPlugins` should already have worked (at least theoretically).

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
```diff
@@ -66,23 +65,15 @@ public function __construct(
...
-        } else {
-            $endpoint = self::NAVIGATION_ENDPOINT;
-        }
-
         $url = sprintf(
             '%s/%s/%s?%s',
             self::BASE_URL,
             $this->shopUrl,
-            $endpoint,
+            static::ENDPOINT,
```
ihanli: Are we sure that `self::ENDPOINT` won't work here?

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
zep-com: Since we have disabled Findologic by default in the last version, it did not get activated by using `ensureLoadedPlugins`'s value after a fresh setup of the environment. Not sure why though.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
zep-com: Yes it was getting the Querybuilder's `index` endpoint, not the child class's overridden endpoint.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
ihanli: Well, that's unfortunate.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder.php
ihanli: I checked the PHP manual (just because I was curious). It seems that constants are definitely not statics but you have to access it this way since the constant is shared between multiple classes (QueryBuilder and Navigation-/SearchQueryBuilder).

The more you know.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/NavigationQueryBuilderTest.php
ihanli: Well, that's unfortunate but okay.

PR #211, opened 2019-04-16
SW-257 legacy export compatibility

zep-com:


FinSearchUnified/BusinessLogic/ExportFactory.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    private $installerService;
+
+    private $decorators = [
+        'fin_search_unified.article_model_factory',
+        'fin_search_unified.shopware_process'
+    ];
+
+    /**
+     * ExportFactory constructor.
```
ihanli: Personally not a fan of this. It should already be clear that `__contruct` is the constructor for this class. Please remove.

PR #214, opened 2019-04-25
SW-261 support multiple range sliders

zep-com:


FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -593,4 +595,181 @@ public function testParsesSmartDidYouMeanData(
...
+     * @param bool $expectedFacetResult
+     * @param string $expectedMinField
+     * @param string $expectedMaxField
+     *
+     * @throws Zend_Http_Client_Exception
+     */
+    public function testCreateRangeSliderFacetMethod(
+        array $filterData,
+        array $parameters,
+        $expectedFacetResult,
```
ihanli: Rename this to `$expectedFacetState` since it is a boolean and not an actual facet.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -533,8 +533,19 @@ private static function createRangeSlideFacet(array $facetItem, $min, $max, $act
...
+        $maxFieldName = 'max' . $facetItem['name'];
+
+        if ($facetItem['name'] === 'price') {
+            $minFieldName = 'min';
+            $maxFieldName = 'max';
+        }
+
+        if ($request->getParam($minFieldName) === null && $request->getParam($maxFieldName) === null) {
+            $active = false;
+            // Perform a loose comparison here since the floating numbers could actually be integers.
```
ihanli: The idea was that this `if` block is exactly structured like this:
```
if (is null) {
    $active = false;
}
// Perform a loose comparison since the floating numbers could actually be integers.
elseif (values match) {
    $active = true;
} else {
    $active = false
}
```
This way it will be clear where the comment belongs to. If this raises a code style error, I would propose to move the comment inside the `elseif`.

PR #215, opened 2019-04-26
SW-266-dont-handle-emotion-banners

zep-com:


FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -336,6 +336,27 @@ public function testUseShopSearchWhenRequestIsNull()
...
+            ->disableOriginalConstructor()
+            ->getMock();
+        $front->expects($this->atLeastOnce())
+            ->method('Request')
+            ->willReturn($request);
+
+        // Assign mocked session variable to application container
+        Shopware()->Container()->set('front', $front);
+
+        $result = StaticHelper::useShopSearch();
```
ihanli: Session variable `findologicDI` should also be set, just to be more explicit. Do the following:
* Set session variable `findologicDI` to `false` inside this test.
* Delete the session variable in method `tearDown`.

PR #216, opened 2019-04-26
SW-263 dont add local changes to archive

zep-com:


archive.sh
```diff
@@ -11,6 +11,24 @@ VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plug
...
+
+git fetch --all --tags --prune
+
+TAG=$(git tag -l "v${VERSION}")
+
+# If tag is available we proceed with the checkout and zip commands
+# else exit with code 1
+if [[ -z "${TAG}" ]]
+then
+echo "[ERROR]: Tag v${TAG} not found"
```
ihanli: Add indentations so it looks more like
```
if
then
    echo
    exit
else
    echo
fi
```

archive.sh
```diff
@@ -11,6 +11,24 @@ VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plug
...
+exit 1
+else
+echo "Checking out ${TAG} from master .."
+fi
+
+git checkout tags/${TAG} -b master
```
ihanli: * I'm not sure if this comment is correct. Wouldn't it create a new branch `master` having `tags/${TAG}` as base?
* Wouldn't `git checkout v${TAG}` do the right thing?

archive.sh
```diff
@@ -27,4 +45,10 @@ cd /tmp
 zip -r9 ${ROOT_DIR}/FinSearchUnified-${VERSION}.zip FinSearchUnified -x FinSearchUnified/phpunit.xml.dist FinSearchUnified/shopware-phpcs.xml FinSearchUnified/Tests/\*
 
 # Delete the directory after script execution
-rm -rf "/tmp/FinSearchUnified"
\ No newline at end of file
+rm -rf "/tmp/FinSearchUnified"
+
+cd ${ROOT_DIR}
+
+git checkout master
```
ihanli: Print a message before this line which says `Restoring work in progress ... `

archive.sh
```diff
@@ -11,6 +11,27 @@ VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plug
...
+BRANCH=$(git branch | grep \* | cut -d ' ' -f2)
+
+# If tag is available we proceed with the checkout and zip commands
+# else exit with code 1
+if [[ -z "${TAG}" ]]
+then
+    echo "[ERROR]: Tag ${TAG} not found"
+    exit 1
+else
+    echo "Checking out ${TAG} from ${BRANCH} .."
```
ihanli: There is a missing `.`. Every other message has three dots while this one only has two.

archive.sh
```diff
@@ -24,7 +45,16 @@ composer install --no-dev
 cd /tmp
 
 # Run archive command to create the zip in the root directory
-zip -r9 ${ROOT_DIR}/FinSearchUnified-${VERSION}.zip FinSearchUnified -x FinSearchUnified/phpunit.xml.dist FinSearchUnified/shopware-phpcs.xml FinSearchUnified/Tests/\*
+zip -r9 ${ROOT_DIR}/FinSearchUnified-${VERSION}.zip FinSearchUnified -x FinSearchUnified/phpunit.xml.dist
+FinSearchUnified/shopware-phpcs.xml FinSearchUnified/Tests/\*
```
ihanli: This doesn't work. `shopware-phpcs.xml` and the `Tests` folder are in the archive now. I guess you will have to do something like
```
zip -r9 ${ROOT_DIR}/FinSearchUnified-${VERSION}.zip FinSearchUnified -x FinSearchUnified/phpunit.xml.dist \
FinSearchUnified/shopware-phpcs.xml FinSearchUnified/Tests/\*
```

archive.sh
```diff
@@ -11,6 +11,27 @@ VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plug
 VERSION="$(echo -e "${VERSION_RAW}" | tr -d '[:space:]')"
 echo "Version: ${VERSION}"
 
+echo $(git stash)
```
ihanli: I found a bug while click testing where this "causes an error" when there were no changes to be stashed. I would propose following approach:
1. Run `git stash` and save the return value.
2. Print the return value which should actually be a message.
3. Run `git stash pop` only if the previously saved return value equals `No local changes to save`.

zep-com: @ihanli I hope you mean when the return value does **NOT** equal to the the provided message else we don't need to apply the stash

ihanli: @zep-com 
Yes, that's what I meant.

archive.sh
```diff
@@ -11,6 +11,29 @@ VERSION_RAW="$(sed -n 's|<version>\(.*\)</version>|\1|p' ./FinSearchUnified/plug
...
+BRANCH=$(git branch | grep \* | cut -d ' ' -f2)
+
+# If tag is available we proceed with the checkout and zip commands
+# else exit with code 1
+if [[ -z "${TAG}" ]]
+then
+    echo "[ERROR]: Tag ${TAG} not found"
+    exit 1
+else
+    echo "Checking out ${TAG} ... "
```
ihanli: Since the branch info was dropped from this message, the else part can be deleted completely since `git checkout` will already print a similar message.

PR #224, opened 2019-05-27
SW-275-fix-live-reloading

zep-com:


FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -69,13 +69,12 @@ public function testBeforeListingCountActionIfFindologicSearchIsActive(
 
         foreach ($requestParameters as $key => $value) {
             $this->assertArrayHasKey($key, $params, sprintf('Expected %s to be present', $key));
```
ihanli: Is this actually necessary if `$this->assertSame($expectedRequestParameters, $params, $expectedMessage);` is done further below? Seems to me that `assertSame` would already make sure that the keys are present.

Please check that and remove the `foreach` (incl. `$requestParameters`) if it is indeed not necessary.

zep-com: Yes it isnt necessary. I had actually removed that but then for the sake of being explicit I kept it the way it was before. But sure will remove it.

PR #226, opened 2019-05-31
Sw 272 implement config cdn

zep-com:


FinSearchUnified/Tests/Components/ConfigLoaderTest.php
```diff
@@ -0,0 +1,425 @@
...
+
+        $responseCallback = $this->returnCallback(function ($url) use ($httpResponse) {
+            Assert::assertSame($this->expectedUrl, $url);
+
+            return $httpResponse;
+        });
+
+        $httpClientMock->expects($this->once())
+            ->method('get')
+            ->will($this->onConsecutiveCalls($responseCallback));
```
ihanli: Use `$responseCallback` together with `->with` and use `->will` to return `$httpResponse`.

PR #230, opened 2019-06-10
SW-269-create-text-facet-handler

zep-com:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
+            'Manufacturer'
+        );
+        $criteria = new Criteria();
+        if ($condition !== null) {
+            $criteria->addCondition($condition);
+        }
+
+        $data = '<?xml version="1.0" encoding="UTF-8"?><searchResult></searchResult>';
+        $filter = new SimpleXMLElement($data);
+        foreach ($filterData as $key => $value) {
```
wladi0097: Tests should be simple, this is way to complex. We need a test to test the test if it is to complex.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
```
wladi0097: Tests should be simple, this is way to complex. We need a test to test the test if it is to complex.

wladi0097: At least move this in its own function called like 'generateFilter()', also you could reuse this new function in line 137.

zep-com: @wladi0097 sure, we can do that. 
thanks

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,147 @@
...
+     */
+    private function getValueListItems(FacetInterface $facet, Criteria $criteria, SimpleXMLElement $filterItems)
+    {
+        $items = [];
+        $actives = [];
+
+        /** @var ProductAttributeFacet $facet */
+        $condition = $criteria->getCondition($facet->getName());
+        if ($condition !== null) {
+            $actives[] = $condition->getValue();
```
ihanli: This will only work if `ProductAttributeFacet::MODE_RADIO_LIST_RESULT` is used. In any other case `$condition->getValue()` will return an array (even if only a single value is possible). Assign the return value of `$condition->getValue()` directly to `$actives` and perform this check afterwards (outside the if condition):
```
if (!is_array($actives)) {
    $actives = [$actives];
}
```

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        foreach ($filterItems as $filterItem) {
+            $name = (string)$filterItem->name;
+            $freq = (int)$filterItem->frequency;
+            if ($index = array_search($name, $actives) === false) {
```
ihanli: 1. Really just my personal preference but add an empty line before the if statement.
2. I'm not sure if this does what it is supposed to. Won't it evaluate `array_search($name, $actives) === false` and assign the return value to `$index`?

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,147 @@
...
+                $active = false;
+            } else {
+                $active = true;
+                array_splice($actives, $index, 1);
```
ihanli: I'm not 100% sure but wouldn't `unset($actives[$index])` do the same?

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
+
+        $this->assertSame($doesSupport, $result);
+    }
+
+    public function filterProvider()
+    {
+        return [
+            'Category filter with "select" type' => ['cat', 'select', false],
+            'Category filter with "label" type' => ['cat', 'label', false],
+            'Vendor filter with "select" type' => ['vendor', 'label', true],
```
ihanli: Guess this should be `select` rather than `label`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
+                    [
+                        new ValueListItem(
+                            'FINDOLOGIC',
+                            'FINDOLOGIC',
+                            true
+                        )
+                    ],
+                    'vendor'
+                ),
+                new ProductAttributeCondition('vendor', ConditionInterface::OPERATOR_EQ, 'FINDOLOGIC'),
```
ihanli: Affects all conditions in this data provider:
Convert `'FINDOLOGIC'` into an array since it also will be like this in production.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    /**
+     * @dataProvider valueListDataProvider
+     *
+     * @param array $filterData
+     * @param ValueListFacetResult $facetResult
+     * @param ConditionInterface|null $condition
+     */
+    public function testFacetModeForListFacetResult(
```
ihanli: I think `testFacetModeForListFacetResult` and `testFacetModeForRadioFacetResult` don't have to be separate tests since they are 99% similar. My suggestion would be:
1. Rename the test to `testGeneratesPartialFacetBasedOnFilterDataAndActiveConditions`.
2. Use `FacetResultInterface` as typehint instead of `ValueListFacetResult` or `RadioFacetResult`.
3. Use `valueListDataProvider` and `radioListDataProvider` for the same test.
4. Pass the mode via the data provider instead of hard coding it.

At the end you should only have `testGeneratesPartialFacetBasedOnFilterDataAndActiveConditions` instead of `testFacetModeForListFacetResult` and `testFacetModeForRadioFacetResult`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,360 @@
...
+        array $filterData,
+        ValueListFacetResult $facetResult,
+        ConditionInterface $condition = null
```
ihanli: Do you know why this works? `null` is definitely not an instance of `ConditionInterface`, so shouldn't this cause an fatal error?

zep-com: it works because we are providing a default value to the interface. If we don't provide the default value and then pass `null` from the data provider, it would throw an error. There is a feature in I think php 7 which typehints a null value but it is not supported in 5.6 so I used the default value instead.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
zep-com: I had done unset, but it was not working for some reason. I will check that again in the review changes

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
zep-com: aha, yes thats evaluating to false. I will fix that

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,154 @@
...
+class TextFacetHandler implements PartialFacetHandlerInterface
+{
+    const ALLOWED_FILTER_TYPES = ['select', 'label'];
+
+    /**
+     * @param FacetInterface $facet
+     * @param Criteria $criteria
+     * @param SimpleXMLElement $filter
+     *
+     * @return RadioFacetResult|ValueListFacetResult|null
```
ihanli: Use `FacetResultInterface|null` instead.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,154 @@
...
+
+        if (!is_array($actives)) {
+            $actives = [$actives];
+        }
+
+        foreach ($filterItems as $filterItem) {
+            $name = (string)$filterItem->name;
+            $freq = (int)$filterItem->frequency;
+            $index = array_search($name, $actives);
+            if ($index === false) {
```
ihanli: Add an empty line above this condition.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -0,0 +1,154 @@
...
+                $active = true;
+                unset($actives[$index]);
+            }
+
+            if ($freq && !$active) {
+                $label = sprintf('%s (%d)', $name, $freq);
+            } else {
+                $label = $name;
+            }
+            $valueListItem = new ValueListItem(
```
ihanli: Add an empty line above this one.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandlerTest.php
```diff
@@ -0,0 +1,339 @@
...
+                        new ValueListItem(
+                            'FINDOLOGIC',
+                            'FINDOLOGIC',
+                            true
+                        )
+                    ],
+                    'vendor'
+                ),
+                ProductAttributeFacet::MODE_RADIO_LIST_RESULT,
+                new ProductAttributeCondition('vendor', ConditionInterface::OPERATOR_EQ, ['FINDOLOGIC']),
```
ihanli: My comment regarding the value array was only meant for `valueListDataProvider`. In this data provider we can keep the single string value since Shopware will also provide us with a single value for radio lists.

PR #231, opened 2019-07-01
SW-284-config-loader-direct-integration

zep-com:


FinSearchUnified/Components/ConfigLoader.php
```diff
@@ -97,7 +98,7 @@ private function getConfigFile()
      */
     private function getCacheKey()
     {
-        return sprintf('%s_%s', self::CACHE_ID, $this->shopkey);
+        return StaticHelper::removeSpecialCharacters(sprintf('%s_%s', self::CACHE_ID, $this->shopkey));
```
ihanli: Was there a specific reason to use `StaticHelper::removeSpecialCharacters`? The only case where you could get an invalid cache key is when you enter a wrong shopkey, which is unlikely to happen (assuming the customer did nothing wrong).

FinSearchUnified/Constants.php
```diff
@@ -19,4 +19,5 @@ class Constants
     const INTEGRATION_TYPE_DI = 'Direct Integration';
     const INTEGRATION_TYPE_API = 'API';
     const SDYM_PARAM_FORCE_QUERY = 'forceOriginalQuery';
+    const INTEGRATION_TYPE = 'IntegrationType';
```
ihanli: Give this constant a more specific name like `CONFIG_KEY_INTEGRATION_TYPE`.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -797,4 +812,73 @@ public function priceFilterProvider()
...
+    /**
+     * @dataProvider directIntegrationProvider
+     *
+     * @param string $integrationType
+     * @param string $expectedResult
+     * @param array $directIntegration
+     *
+     * @throws Exception
+     */
+    public function testDirectIntegration($integrationType, $expectedResult, $directIntegration)
```
ihanli: Add the type hint for `$directIntegration`.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -797,4 +812,73 @@ public function priceFilterProvider()
...
+        $config = ['directIntegration' => ['enabled' => $directIntegration]];
+
+        $mockedCache->expects($this->once())
+            ->method('load')
+            ->willReturn($config);
+
+        $mockConfig->method('offsetGet')->willReturnMap($configArray);
+        $configLoader = new ConfigLoader(
+            $mockedCache,
+            Shopware()->Container()->get('http_client'),
```
ihanli: Use a mocked client so we don't accidentally perform a real HTTP request.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -797,4 +812,73 @@ public function priceFilterProvider()
...
+            $mockConfig
+        );
+
+        Shopware()->Container()->set('fin_search_unified.config_loader', $configLoader);
+        Shopware()->Container()->set('config', $mockConfig);
+
+        $isDirectIntegration = StaticHelper::checkDirectIntegration();
+
+        // Reset this service directly here as it is being overridden only for this test
+        Shopware()->Container()->reset('fin_search_unified.config_loader');
```
ihanli: Do the resetting at the very and of the test.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -797,4 +812,73 @@ public function priceFilterProvider()
...
+        $plugin = $pluginManager->getPluginByName('FinSearchUnified');
+        $config = $pluginManager->getPluginConfig($plugin);
+
+        $this->assertSame($expectedResult, $config['IntegrationType']);
+    }
+
+    public function directIntegrationProvider()
+    {
+        return [
+            'IntegrationType is API and directIntegration is not enabled' => ['API', 'API', false],
```
ihanli: We have to change the wording a little bit:
* `IntegrationType` -> `Integration type`.
* `directIntegration` -> `DI`.
* `IntegrationType is Direct...` -> `Integration type is DI...`.

We can use the acronym `DI` instead of `Direct Integration`. Customers don't know what `DI` is but since we are the only ones to see it, we will keep it easy for us :wink:

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -856,4 +852,14 @@ public static function setSmartDidYouMean(SimpleXMLElement $xmlResponse)
             ]);
         }
     }
+
+    public static function removeSpecialCharacters($name, $replace = '')
```
ihanli: Check my previous comment regarding this method. The use of it shouldn't be necessary unless there is an obvious reason that I'm missing.

FinSearchUnified/Components/ConfigLoader.php
zep-com: Yes apparently we are testing for an invalid shopkey in the test which has spaces so it was failing for the "Findologic Shopkey"

zep-com:
> The Travis builds are failing currently due to linting issues. Please check those too.

Ahan, weird. I had run the lint check. Wonder why it didn't work.

FinSearchUnified/Helper/StaticHelper.php
zep-com: Please see my response, thanks.

FinSearchUnified/Components/ConfigLoader.php
ihanli: Which test is that exactly? I just ran all of them using `return sprintf('%s_%s', self::CACHE_ID, $this->shopkey);` and they passed.

PR #235, opened 2019-07-15
Sw 283 code linting travis build

zep-com:


.travis.yml
```diff
@@ -1,19 +1,18 @@
 language: php
 
 php:
-  - 5.6
-  - 7.0
-  - 7.1
+  - "5.6"
```
ihanli: Why were the double quotes added? Seemed to work without them before.

zep-com: yes it worked without them, but was just following how it was mentioned in the docs. It expects a string rather than a number.

zep-com: Also, I think based on this info, it seems that the number is needed for more precise version such as 7.1.9. Just an assumption. 
[https://docs.travis-ci.com/user/languages/php/#choosing-php-versions-to-test-against](url)

ihanli: Both versions exist in the docs but let's use it with double quotes since I don't have a strong opinion on this and the docs don't state anything specific.

PR #236, opened 2019-07-18
Sw 279 facet handler category filter

zep-com:


FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
```diff
@@ -0,0 +1,166 @@
...
+     * @return array
+     */
+    private function parseCategories(SimpleXMLElement $filterItems, array $actives)
+    {
+        foreach ($filterItems as $filterItem) {
+            $name = (string)$filterItem->name;
+            $frequency = (int)$filterItem->frequency;
+
+            // Only add the items of $filterItems that are not already present in $actives
+            $index = array_search($name, $actives);
```
ihanli: This will not work since `$actives` is a multi dimensional array. You will have to search the array recursively by not only considering the current filter item, but also the parents of it.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
```diff
@@ -0,0 +1,194 @@
...
+
+            if (array_key_exists('frequency', $category) && !$category['active']) {
+                $label = sprintf('%s (%d)', $key, $category['frequency']);
+            }
+
+            $item = new TreeItem(
+                $id,
+                $label,
+                $category['active'],
+                self::getTreeItems($category['children'], $id)
```
ihanli: Use `$this` instead of `self`.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Helper method to check if needle exists in array recursively
+     *
+     * @param mixed $needle
+     * @param array $haystack
+     *
+     * @return bool
+     */
+    private function keyExists($needle, array $haystack)
```
ihanli: Wouldn't this method return `true` for any category that is present in `$haystack`, not only if it's actually active?

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        foreach ($haystack as $key => $value) {
+            if ($key === $needle) {
+                return true;
+            }
+
+            if (is_array($value)) {
+                if ($x = $this->keyExists($key, $value)) {
```
ihanli: The initialization of `$x` shouldn't be necessary.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandlerTest.php
```diff
@@ -0,0 +1,445 @@
...
+        $result = $facetHandler->generatePartialFacet($facet, $criteria, $filter);
+
+        $this->assertEquals($facetResult, $result);
+    }
+
+    public function treeItemProvider()
+    {
+        return [
+            'Facet with filter without condition' => [
+                // $filterData
```
ihanli: You can use array keys for this instead of comments e.g. `filterData => ['name` => 'cat',...]`

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
zep-com: technically yes, but the haystack in this case is only the actives array itself

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
zep-com: yes it was used to return the index initially, but forgot to remove it afterwards. will update that.

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/CategoryFacetHandler.php
ihanli: Ah, okay. The actives have already been parsed at this point. Got it.

PR #238, opened 2019-07-29
SW-288 Add new FINDOLOGIC snippet

TobiasGraml11:


FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -147,13 +147,18 @@ public function onFrontendPostDispatch(Enlight_Event_EventArgs $args)
         $groupKey = Shopware()->Session()->get('sUserGroup', 'EK');
         $hash = StaticHelper::calculateUsergroupHash($this->getShopKey(), $groupKey);
 
+        $searchResultContainer = Shopware()->Config()->get('SearchResultContainer', 'fl-result');
```
ihanli: Remove the default values for both configs. There are already defaults in the `config.xml` and they can't be emptied since they are required.

PR #245, opened 2019-08-08
SW-294-fix-product-export

zaifastafa:


FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -138,4 +144,108 @@ public function testEmptySuppliersAreSkipped(array $articleConfiguration)
...
+                            ],
+                        ]
+                    ],
+                    'attribute' => [
+                        'attr1' => 'Article Attribute'
+                    ]
+                ],
+                'Article Attribute'
+            ],
+            'Variant Attribute' => [
```
ihanli: Something seems odd about this data provider: Shouldn't this data set actually fail because for Shopware 5.5 we try to get the attributes from the article not its main variation?

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -715,8 +720,12 @@ protected function setProperties()
             $allProperties[] = new Property('release_date', ['' => $releaseDate]);
         }
 
-        /** @var \Shopware\Models\Attribute\Article $attributes */
-        $attributes = $this->baseArticle->getAttribute();
+        if (is_callable([$this->baseArticle, 'getAttribute']) && !is_null($this->baseArticle->getAttribute())) {
```
ihanli: * Why the additional check for `is_null`?
* The requested comment explaining why this check has to be done, is missing.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
ihanli: Check why the test doesn't fail with this data set. It might be valid but we should still know the reason. Post your answer here.

zaifastafa: I checked the code, its because if there is an attribute on the root article, it assigns it in the mainDetail article attribute and then unsets the root article attribute index. So basically it is getting the attribute data from the main detail as the root article attribute is always null.
Thats just my observation. If you have some insight do let me know too.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
zaifastafa: Based on my previous comment above, the getAttribute in the Article is always returning null as it assigns it to the mainDetail, and hence the `attr1` is never set in the Article attribute, but always in the Detail.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
ihanli: Interesting....It works in production at least. Let me do some checks.

PR #248, opened 2019-08-12
SW-301 Fix warnings from Shopware store

TobiasGraml11:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -121,7 +121,7 @@ protected static function redirectOnLandingpage(\SimpleXMLElement $xmlResponse)
      */
     protected static function setFallbackFlag($status)
     {
-        setcookie('Fallback', $status);
+        setcookie('Fallback', $status, 0, "", "", true);
```
TheKeymaster: Use single quotes instead of double quotes.

PR #250, opened 2019-08-23
SW-297 whitelist checkout requests

zaifastafa:


FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -219,4 +233,28 @@ private function isManufacturerPage(Request $request)
...
+    /**
+     * @param Request $request
+     *
+     * @return bool
+     */
+    public function isWhiteListed(Enlight_Controller_Request_Request $request)
+    {
+        $controllerName = strtolower($request->getControllerName());
+
+        foreach (self::WHITE_LIST as $key => $value) {
```
TheKeymaster: Rename `$key` to `$name` and `$value` to `$actions`.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -219,4 +233,28 @@ private function isManufacturerPage(Request $request)
     {
         return $request->getControllerName() === 'listing' && $request->getActionName() === 'manufacturer';
     }
+
+    /**
+     * @param Request $request
```
TheKeymaster: The docblock and the type hint do not match. The actual type is `Enlight_Controller_Request_Request` and not `Request`.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -219,4 +233,28 @@ private function isManufacturerPage(Request $request)
...
+    public function isWhiteListed(Enlight_Controller_Request_Request $request)
+    {
+        $controllerName = strtolower($request->getControllerName());
+
+        foreach (self::WHITE_LIST as $key => $value) {
+            if (!is_array($value)) {
+                if ($controllerName === $value) {
+                    return true;
+                }
+            } else {
```
TheKeymaster: Make this an `elseif` and move the conditions to it.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -150,20 +172,19 @@ public function onFrontendPostDispatch(Enlight_Event_EventArgs $args)
         $searchResultContainer = Shopware()->Config()->get('SearchResultContainer');
         $navigationContainer = Shopware()->Config()->get('NavigationContainer');
 
-        try {
```
ihanli: This try/catch should stay too.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -190,33 +219,55 @@ public function onAjaxSearchIndexAction()
...
+     * @param Enlight_Controller_Request_Request $request
+     *
+     * @return bool
+     */
+    public function isWhiteListed(Enlight_Controller_Request_Request $request)
+    {
+        $controllerName = strtolower($request->getControllerName());
+
+        foreach (self::WHITE_LIST as $name => $actions) {
+            if (!is_array($actions)) {
```
ihanli: I think these condition checks can be optimized:
```PHP
return $controllerName === $actions || ($controllerName === $name && in_array($request->getActionName(), $actions));
```
Also add a comment explaining the conditions.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -49,22 +64,21 @@ public static function getSubscribedEvents()
...
-        $controllerName = strtolower($request->getControllerName());
-
         if ($this->isSearchPage($request)) {
             Shopware()->Session()->offsetSet('isSearchPage', true);
             Shopware()->Session()->offsetSet('isCategoryPage', false);
         } elseif ($this->isCategoryPage($request)) {
             Shopware()->Session()->offsetSet('isCategoryPage', true);
             Shopware()->Session()->offsetSet('isSearchPage', false);
-        } elseif ($this->isManufacturerPage($request) || !(in_array($controllerName, self::WHITE_LIST, true))
+        } elseif ($this->isManufacturerPage($request) || !$this->isWhiteListed($request)
```
ihanli: Since this is shorter now, maybe the `) {` would fit on the same line?

FinSearchUnified/Subscriber/Frontend.php
zaifastafa: There was no Enlight_Exception being thrown in the code. Should I change it to a Exception instead?

ihanli: Yes, please. In any error case this shouldn't break the shop but fail silently.

FinSearchUnified/Subscriber/Frontend.php
zaifastafa: Simplifying this wont work as we need to check if any value returns true. Simplifying this would also return the first false encountered which is not intended behavior.

ihanli: You are right. I forgot about the loop. How about
```PHP
if ($controllerName === $actions || ($controllerName === $name && in_array($request->getActionName(), $actions))) {
    return true;
}
```
?

zaifastafa: ah, yes cant argue with this  :wink:

ihanli:
@TheKeymaster 
Please provide your feedback or approve the changes (since the issues) were fixed.

PR #251, opened 2019-08-24
Sw 302 encode product url

zaifastafa:


FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -247,4 +247,111 @@ public function articlesWithAttributeProvider()
...
+     * @throws ReflectionException
+     */
+    public function testArticleWithSEOUrl(array $articleConfiguration, $expectedUrl)
+    {
+        $baseCategory = new Category();
+        $baseCategory->setId(5);
+
+        $articleFromConfiguration = $this->createTestProduct($articleConfiguration);
+
+        $shop = Manager::getResource('Shop')->getRepository()->find(1);
```
ihanli: Are these two lines actually necessary? You would think that the main shop is already registered.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -247,4 +247,111 @@ public function articlesWithAttributeProvider()
...
+        $xmlArticle = $findologicArticle->getXmlRepresentation();
+
+        $reflector = new ReflectionClass(Item::class);
+        $properties = $reflector->getProperty('url');
+        $properties->setAccessible(true);
+        $values = $properties->getValue($xmlArticle);
+        $actualUrl = current($values->getValues());
+
+        // Remove the host from the URL for assertion
+        $actualUrl = explode(Shopware()->Shop()->getHost(), $actualUrl)[1];
```
ihanli: I don't have a good feeling about this. Since we are adding the shop's URL to the product URL it might happen that we are doing something wrong. In that case we would never notice it because we don't consider the shop's hostname in the test.
What if we added `Shopware()->Shop()->getHost()` to `$expectedUrl` instead of removing it from `$actualUrl`?

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
zaifastafa: Apparently the conversion does not work without re-registering the shop. If you remove these two lines, it returns the default URL with the `detail/index/sArticle/1` type of format. It seems that there are some other services needed to registered again for the rewrite URL to work.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -247,4 +247,110 @@ public function articlesWithAttributeProvider()
...
+                        'inStock' => 16,
+                        'prices' => [
+                            [
+                                'customerGroupKey' => 'EK',
+                                'price' => 99.34,
+                            ],
+                        ]
+                    ],
+                ],
+                sprintf('http://%s%s', $host, '/abdr%C3%BCckklotz-f%C3%BCr%2Bbutler%20reifenmontierger%C3%A4t%2F')
```
ihanli: Not entirely wrong but the third parameter could be part of the first one e.g. `sprintf('http://%s/abdr%%C3%%BCckklotz-f%%C3%%BCr%%2Bbutler%%20reifenmontierger%%C3%%A4t%%2F', $host)`.

Don't forget to update `sprintf('http://%s%s', $host, '/reifenmontage%2F')` too.

zaifastafa: Doing that gives a warning as it tries to check for a parameter for replacement due to `%2F`. If its not an issue let me know I will change it

ihanli: Have you used `%2F` or `%%2F`? Because the latter version should escape the % character and actually prevent that issue.

zaifastafa: ahan I see. didnt know that. will push the changes in a bit

ihanli: I made the same mistake on my first day at FINDOLOGIC. That's why I know how to escape it  :sweat_smile:

PR #252, opened 2019-08-26
SW-298-handle-special-characters-in-keywords

zaifastafa:


FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -247,4 +247,63 @@ public function articlesWithAttributeProvider()
...
+        $reflector = new ReflectionClass(Item::class);
+        $properties = $reflector->getProperty('keywords');
+        $properties->setAccessible(true);
+        $keywords = $properties->getValue($xmlArticle);
+        $values = $keywords->getValues();
+        $this->assertNotEmpty($values);
+
+        foreach ($values as $value) {
+            foreach ($value as $item) {
+                $this->assertTrue(in_array($item->getValue(), $expectedKeywords));
```
ihanli: Imagine following scenario:
`$expectedKeywords = ["I'm a simple string", "\xC2\xBD"]` and `$value = ["I'm a simple string"]`

The test would not fail although we are missing some data. This assertion needs to be rewritten to make sure that we are not missing anything.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -247,4 +247,63 @@ public function articlesWithAttributeProvider()
...
+                    [
+                        'customerGroupKey' => 'EK',
+                        'price' => 99.34,
+                    ],
+                ]
+            ],
+            'keywords' => "I'm a simple string,\xC2\xBD,\xC2\x80"
+        ];
+
+        $expectedKeywords = ["I'm a simple string", "\xC2\xBD"];
```
ihanli: Apparently Shopware converts `\xC2\x80` to `\xC3\x82\xC2\x80`. So we end up with `$expectedKeywords` being `["I'm a simple string", "\xC2\xBD", "\xC3\x82"]`. While the last entry makes no sense as a keyword, it is still a valid character.

zaifastafa: I didn't understand this. How did we know its breaking it into two characters?

ihanli: I saw it during the click test. After Shopware has loaded the keywords it becomes something like `A[]` and after removing the control character all that's left is `A`.

zaifastafa: @ihanli So should we do the assertion for this extra character or replace it with some other one? because that seems like a garbage character which we have no use of and we are not even setting it technically, so why should we "expect" this character as a keyword?

ihanli: Even if it makes no sense we will end up in the same situation on production, so we are just replicating the scenario.
I would suggest we either add that one character to the expected keywords or use a different one which can be stripped completely. The latter options makes more sense in my opinion but I will leave the decision to you.

zaifastafa: @ihanli I am not getting this extra character at my setup. 
image

ihanli: okay ... :confused: 
Interesting, I can't reproduce it anymore. I will resolve this conversation, so the only thing left is the issue with the assertions.

PR #254, opened 2019-08-26
SW-299 Add redirect function for old urls

TobiasGraml11:


FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -56,20 +56,18 @@ public function onFrontendPreDispatch(Enlight_Event_EventArgs $args)
...
-            Shopware()->Session()->offsetSet('isCategoryPage', false);
-        } elseif ($this->isCategoryPage($request)) {
-            Shopware()->Session()->offsetSet('isCategoryPage', true);
-            Shopware()->Session()->offsetSet('isSearchPage', false);
-        } elseif ($this->isManufacturerPage($request) || !(in_array($controllerName, self::WHITE_LIST, true))
-        ) {
-            Shopware()->Session()->offsetSet('isCategoryPage', false);
-            Shopware()->Session()->offsetSet('isSearchPage', false);
+        if ($this->isLegacySearch($request)) {
+            $params = $request->getQuery();
```
ihanli: You definitely need to add more whitespace, otherwise you will end up with a giant blob of code. In comparison:
```
            $params = $request->getQuery();

            unset($params['module']);
            unset($params['controller']);
            unset($params['action']);

            $url = '/search?' . http_build_query($params, null, '&', PHP_QUERY_RFC3986);

            // Perform a redirect to the actual search path and tell the client that it moved permanently.
            // The legacy parameters (relevant for search) will be mapped automatically by the corresponding
            // event handler.
            $subject->redirect($url, ['code' => 301]);
```
This also exactly how the code was provided in the story.

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -219,4 +217,42 @@ private function isManufacturerPage(Request $request)
...
+    /**
+     * @param Request $request
+     *
+     * @return bool
+     */
+    private function isLegacySearch(Request $request)
+    {
+        $params = $request->getQuery();
+
+        if ($params['controller'] === 'FinSearchAPI' && $params['action'] === 'search') {
```
ihanli: `return $params['controller'] === 'FinSearchAPI' && $params['action'] === 'search'` will do the same.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -79,6 +79,68 @@ public function vendorProvider()
         ];
     }
 
+    public function legacyUrlProvider()
+    {
+        return [
+            'Normal search query' => [
```
ihanli: Try to avoid words like "normal" since that could mean anything. Maybe `Query without filters`?

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -79,6 +79,68 @@ public function vendorProvider()
...
+            ],
+            'Query with filter and special characters' => [
+                'params' => [
+                    'controller' => 'FinSearchAPI',
+                    'action' => 'search',
+                    'sSearch' => 'test',
+                    'attrib' => ['vendor' => ['Shopware / Österreich#%']],
+                ],
+                'expectedUrl' => '/search?' . http_build_query([
+                        'sSearch' => 'test',
```
ihanli: The indentation needs to fixed here.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -234,6 +296,47 @@ public function testBeforeSearchIndexAction($vendor)
...
+            ->setMethods(['Request', 'redirect'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $subject->method('Request')
+            ->willReturn($request);
+
+        if ($expectedUrl === null) {
+            $subject->expects($this->never())->method('redirect');
+        } else {
+            $subject->method('redirect')->with($expectedUrl, ['code' => 301]);
```
ihanli: Assert a call count here.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -234,6 +296,47 @@ public function testBeforeSearchIndexAction($vendor)
...
+        }
+
+        // Create mocked args for getting Subject and Request
+        $args = $this->getMockBuilder(Enlight_Event_EventArgs::class)
```
ihanli: * Generally speaking: You should prefer the actual implementation over mocks. Sometimes it's not that easy to get an actual instance (e.g. because of dependencies), in which case we have to settle for a mock.
* In this case: `Enlight_Controller_Action` and `Enlight_Event_EventArgs` can be used directly and don't have to be mocked.
* Check `FrontendTest::testSessionValuesAfterListingCount`.

PR #256, opened 2019-09-14
Sw 291 using less files

zaifastafa:


FinSearchUnified/Tests/FinSearchUnifiedTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * @dataProvider customPluginProvider
+     *
+     * @param bool $isActive
+     */
+    public function testPluginDeactivateWithCustomPlugin($isActive)
+    {
+        // We use a custom class created below to replicate the custom plugin functionality
+        $plugin = new ExtendFinSearchUnified();
```
TheKeymaster: I do not think that we do not need to create a separate class below. Try to create an instance of `\Shopware\Models\Plugin\Plugin` and set the plugins name accordingly.

FinSearchUnified/Tests/FinSearchUnifiedTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @dataProvider customPluginProvider
+     *
+     * @param bool $isActive
+     */
+    public function testPluginDeactivateWithCustomPlugin($isActive)
+    {
+        // We use a custom class created below to replicate the custom plugin functionality
+        $plugin = new Shopware\Models\Plugin\Plugin();
```
TheKeymaster: Import this class as `PluginModel`.

PR #258, opened 2019-09-17
SW-258-export-properties-attributes

zaifastafa:
One test fails due to the legacy attribute


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -406,18 +406,23 @@ protected static function checkIfHasValue($value)
...
-            ']' => '%5D'
-        ];
+        $articleMainImages = [];
+        $replacements = [];
+        $baseLink = Shopware()->Modules()->Core()->sRewriteLink();
+        $mediaService = Shopware()->Container()->get('shopware_media.media_service');
+
+        if ($this->baseArticle->getImages() !== null) {
+            $articleMainImages = $this->baseArticle->getImages()->getValues();
+            $replacements = [
```
TheKeymaster: I think you could declare `$replacements` with these values above. They are static and won't change.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -656,7 +664,7 @@ protected function setAttributes()
         $allAttributes[] = $xmlNewFlag;
 
         // Add free_shipping
-        if ($this->baseVariant->getShippingFree() == '') {
+        if ($this->baseVariant->getShippingFree() === '') {
```
TheKeymaster: Checking the type strictly here may cause issues, because `getShippingFree` will return an `int`.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -740,7 +751,7 @@ protected function setProperties()
         // Since Shopware 5.3, attributes are assigned to the articles main details. Already existing ones aren't
         // touched.
         // \Shopware\Models\Article\Article::getAttribute has been removed in Shopware 5.6 entirely.
-        if (is_callable([$this->baseArticle, 'getAttribute']) && !is_null($this->baseArticle->getAttribute())) {
+        if (is_callable([$this->baseArticle, 'getAttribute']) && $this->baseArticle->getAttribute() !== null) {
```
TheKeymaster: Since we are exporting the free-text fields now as attributes, we do no longer need them as properties.

TheKeymaster:
@zaifastafa no we won't need those test cases anymore, since we removed the code part related to it.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -796,6 +776,54 @@ protected function setProperties()
         }
     }
 
+    /**
+     * @param \Shopware\Models\Attribute\Article $attributes
+     *
+     * @return array
+     */
+    protected function getAttributesByInstance(\Shopware\Models\Attribute\Article $attributes = null)
```
TheKeymaster: Can we import this class? Maybe as `ArticleAttribute`?

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -351,13 +352,13 @@ protected function setUrls()
 
         // This will only encode parts of the URL path and leave separator itself untouched.
         $seoUrl = $shopUrl . array_reduce(explode('/', $urlPath), function ($encodedPath, $item) {
-            $encodedPath .= '/';
+                $encodedPath .= '/';
```
TheKeymaster: Why was this line indent when two lines below it isn't? They should be on the same indent level.

zaifastafa: Not sure, I simply run the composer fix before pushing the changes. Will check it again. Because a similar indentation issue had causing the lint check to fail previously.

zaifastafa: apparently this is the same code block which caused the lint to fail previously. The lint fix does this indentation by itself.

TheKeymaster: I do not know what issue you had with this indent. I just fixed it by reverting the indent change you made on that line ;).

zaifastafa: @TheKeymaster Yes I saw, weird. But did you run `composer fix`? Because if you do, you might  see that the indentation is applied again.

TheKeymaster: @zaifastafa no on my machine nothing is "fixed".
Selection_009

PR #262, opened 2019-10-09
Sw 264 handle search for cached pages

zaifastafa:


FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -3,11 +3,35 @@
...
+
+    /**
+     * @var RewriteMatcher
+     */
+    private $rewrite;
+
+    public function __construct(Zend_Cache_Core $cache, RewriteMatcher $rewrite)
+    {
+        $this->cache = $cache;
+        $this->rewrite = $rewrite;
```
TheKeymaster: Rename this to `rewriteMatcher` as `$rewrite` does not tell what it is or what it does.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -16,14 +40,73 @@ class Widgets implements SubscriberInterface
...
+    public function onWidgetsPreDispatch(Enlight_Event_EventArgs $args)
+    {
+        /** @var Enlight_Controller_Request_RequestHttp $request */
+        $request = $args->get('request');
+        $url = $this->parseReferUrl($request);
+
+        if (strpos($url, 'search') !== false) {
+            Shopware()->Session()->isSearchPage = true;
+            Shopware()->Session()->isCategoryPage = false;
+        } else {
```
TheKeymaster: Consider splitting up this huge else block into multiple smaller functions. This way devs who read this code the first time can see semantics by only reading the method names instead of checking the entire function just to know what a specific code part actually does.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -16,14 +40,73 @@ class Widgets implements SubscriberInterface
...
+            // If the URL is not a search page we will set the session value here
+            Shopware()->Session()->isSearchPage = false;
+
+            // Check if the category page is cached and set session value to true if it is
+            $cacheKey = md5($url);
+            $isCached = $this->cache->load($cacheKey);
+            if ($isCached !== null && $isCached !== false) {
```
TheKeymaster: I think checking for truthiness is enough in that case.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -33,4 +116,25 @@ public function beforeListingCountAction(Enlight_Hook_HookArgs $args)
...
         }
     }
+
+    /**
+     * @param Enlight_Controller_Request_RequestHttp $request
+     *
+     * @return string
+     * @throws Exception
+     */
+    private function parseReferUrl($request)
```
TheKeymaster: Rename this function to `parseReferer` as you already know by the name "referer" that it is an URL.

FinSearchUnified/Tests/Subscriber/FrontendTest.php
```diff
@@ -329,26 +329,36 @@ public function testBeforeSearchIndexAction($vendor)
...
+
         $subject = $this->getMockBuilder(Enlight_Controller_Action::class)
-            ->setMethods(['Request', 'redirect'])
             ->disableOriginalConstructor()
             ->getMock();
-        $subject->method('Request')
-            ->willReturn($request);
 
-        $response = new Enlight_Controller_Response_ResponseHttp();
+        $response = $this->createMock(Enlight_Controller_Response_ResponseHttp::class);
```
TheKeymaster: :+1: nice!

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -19,9 +24,8 @@ public static function setUpBeforeClass()
 
     protected function tearDown()
     {
+        unset(Shopware()->Session()->isSearchPage, Shopware()->Session()->isCategoryPage);
         parent::tearDown();
-
-        Shopware()->Session()->offsetUnset('isSearchPage');
```
TheKeymaster: Unfortunately `offsetUnset` does not return `$this`, so you can not chain those statements. How sad.

FinSearchUnified/Tests/Subscriber/WidgetsTest.php
```diff
@@ -116,4 +120,241 @@ public function searchParameterProvider()
...
+     * @dataProvider categoryPageProvider
+     *
+     * @param string $referer
+     * @param bool $expectedIsCategoryPage
+     * @param string $basePath
+     *
+     * @throws ReflectionException
+     * @throws Zend_Cache_Exception
+     */
+
```
TheKeymaster: Empty line.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -33,4 +116,25 @@ public function beforeListingCountAction(Enlight_Hook_HookArgs $args)
...
+    {
+        $referrer = isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : '';
```
TheKeymaster: I do not really like using superglobals directly. How about `$request->headers->get('referer')`?

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -16,14 +40,73 @@ class Widgets implements SubscriberInterface
...
+            if ($isCached !== null && $isCached !== false) {
+                Shopware()->Session()->isCategoryPage = true;
+
+                return;
+            }
+            // Unless we have a category page, this session variable will always be false
+            Shopware()->Session()->isCategoryPage = false;
+
+            $context = Context::createFromShop(
+                Shopware()->Container()->get('shop'),
```
TheKeymaster: Consider typehinting dynamically loaded instances.

FinSearchUnified/Subscriber/Widgets.php
zaifastafa: ahan I see. Just out of curiosity, what is the difference between `getHeader` and calling it like this? because I think getHeader was returning null so I changed it to super global. I wasn't aware of this property.

zaifastafa: I guess this seems to be failing on travis.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -33,4 +93,61 @@ public function beforeListingCountAction(Enlight_Hook_HookArgs $args)
...
+            $basePath = rtrim($basePath, '/') . '/';
+            $url = str_replace($basePath, '', $url);
+        }
+        $url = ltrim($url, '/');
+
+        return $url;
+    }
+
+    /**
+     * Method to match the given URL if it is a category page and set the isCategoryPage session variable accordingly
```
TheKeymaster: Try to avoid things like `Method` in a docblock as you do not want to know what something is, but rather what it does.

FinSearchUnified/Subscriber/Widgets.php
```diff
@@ -33,4 +93,61 @@ public function beforeListingCountAction(Enlight_Hook_HookArgs $args)
...
+     *
+     * @param string $url
+     */
+    public function matchCategoryPageUrl(string $url)
```
TheKeymaster: Consider returning a boolean here. This way the method could be renamed to `isCategoryPage`. The user of this method just sets `Shopware()->Session()->isCategoryPage = $this->isCategoryPage()`. This not only good for semantic purposes but also reduces side effects of this method.

FinSearchUnified/Subscriber/Widgets.php
TheKeymaster: Im fine with it.

FinSearchUnified/Subscriber/Widgets.php
TheKeymaster: Try to avoid things like `Method` in a docblock as you do not want to know what something is, but rather what it does.

FinSearchUnified/Subscriber/Widgets.php
TheKeymaster: Consider returning a boolean here. This way the method could be renamed to `isCategoryPage`. The user of this method just sets `Shopware()->Session()->isCategoryPage = $this->isCategoryPage()`. This not only good for semantic purposes but also reduces side effects of this method.

PR #271, opened 2019-11-06
Sw 305 mark selected facet xml response

zaifastafa:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -161,17 +165,30 @@ protected function createFacets(Criteria $criteria, SimpleXMLElement $filters)
...
+        /** @var  $criteriaFacet */
         foreach ($criteria->getFacets() as $criteriaFacet) {
             $field = $criteriaFacet->getField();
-            $filter = $filters->xpath(sprintf('//name[.="%s"]/parent::*', $field));
-            if (empty($filter)) {
-                continue;
+            $selectedFilter = $filters->xpath(sprintf('//name[.="%s"]/parent::*', $field));
+
+            if (empty($selectedFilter)) {
+                if ($criteria->hasUserCondition('price')) {
```
TheKeymaster: The story description states

> Check if the criteria object has a condition for the current facet

But here you are only checking for `price`. Instead you should check for `$criteriaFacet->getName()`.

> Pass the facet's name or the facet's field name if it's value is price.

Basically does not matter because the name will always be `price`.

Also I debugged a lot since the implementation does not work as intended. In `\FinSearchUnified\Bundle\SearchBundleFindologic\FacetHandler\TextFacetHandler::createValueListFacetResult` these lines were added:

```php
        if (!$filter->items->item) {
            return null;
        }
```

Not sure why there were added, but this causes the selected filters to be automatically deselected if they have no items.

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
```diff
@@ -264,6 +269,96 @@ public function testProductNumberSearchResultWithAllFilters(SimpleXMLElement $xm
         }
     }
 
+    public function selectedFiltersProvider()
```
TheKeymaster: Since we had some issues that caused selected filters not to work, add a test where the facet is selected but has no items.

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
```diff
@@ -264,6 +269,96 @@ public function testProductNumberSearchResultWithAllFilters(SimpleXMLElement $xm
...
+    {
+        $xmlResponse = $this->getXmlResponse();
+        $filters = $xmlResponse->addChild('filters');
+
+        $this->setPriceFilter($filters);
+
+        return [
```
TheKeymaster: Can we broaden this a little bit? Right now only the price filter is tested. Let's also make sure that the other facet results are tested as well.

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -180,4 +197,51 @@ protected function createFacets(Criteria $criteria, SimpleXMLElement $filters)
...
+
+    /**
+     * @param FacetInterface $facet
+     * @param ConditionInterface $condition
+     *
+     * @return SimpleXMLElement
+     */
+    private function createSelectedFilter(FacetInterface $facet, ConditionInterface $condition)
+    {
+        $data = '<?xml version="1.0" encoding="UTF-8"?><searchResult></searchResult>';
```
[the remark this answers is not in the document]
zaifastafa: @TheKeymaster No it cannot be created without the argument. The `$data` parameter is required. And I added that as root element as we have that in the XML response.

FinSearchUnified/Bundle/ProductNumberSearch.php
zaifastafa: > Not sure why there were added, but this causes the selected filters to be automatically deselected if they have no items.

This was added because the story before this was failing as it was required that there were no items being passed in the filter. I had mentioned the reason for this in the youtrack comment or maybe in the git comment I dont remember properly, but I had tagged @dblaichinger with it. 

Also please elaborate this comment, it is confusing what exactly needs to be done here, whether it is `price` that needs to be passed or `facetName()` ?

zaifastafa: @TheKeymaster In regards to why it was added was: 

> `$values = $this->getValueListItems($facet, $criteria, $filter->items->item);`

This line used to throws `null` error as the `$filter->items->item` does not exist in one of the test scenarios. I think you have given a similar scenario in the comment below as well, so please let me know how is that supposed to be handled if it is not working as intended.

FinSearchUnified/Bundle/ProductNumberSearch.php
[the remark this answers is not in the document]
TheKeymaster: @zaifastafa you are right, a SimpleXMLElement can not be created without arguments. I would suggest renaming the root element to be just `<filter />`.

FinSearchUnified/Bundle/ProductNumberSearch.php
TheKeymaster: @zaifastafa 
> Also please elaborate this comment, it is confusing what exactly needs to be done here, whether it is price that needs to be passed or facetName() ?

This is not price filter related. It should be done for all filters as we need to know weither the filter is available in the XML response. If it is not we ask Shopware if this filter was selected by calling `$criteria->hasCondition($criteriaFacet->getName())`. The price filter does not need to be handled differently then the others in this code path.

> `$values = $this->getValueListItems($facet, $criteria, $filter->items->item);` This line used to throws null error as the $filter->items->item does not exist in one of the test scenarios. I think you have given a similar scenario in the comment below as well, so please let me know how is that supposed to be handled if it is not working as intended.

Removing the filter completely is not the solution to this issue, since it is legitimate that a filter does not have any more items (e.g. it was selected).
Instead create a variable called `$items`, which holds all items (`$filter->items->item`). If the XML path does not exist, then simply initialize it with `null`.
Later in `\FinSearchUnified\Bundle\SearchBundleFindologic\FacetHandler\TextFacetHandler::getValueListItems` when we go through all `$filterItems`, allow the variable to also be `null` and add the required changes to the method so submitting `null` does not throw any errors.
We need to do that because we need to tell Shopware which values are currently selected.

zaifastafa: @TheKeymaster OK it makes sense now. `$criteria->hasCondition($criteriaFacet->getName())` So what does this sentence in the description mean? 

> Pass the facet's name or the facet's field name if it's value is price.

zaifastafa: @TheKeymaster also, can you describe a bit what is the logic of / difference between "selected filters", "facets" and "filter items". How are they interconnected?

TheKeymaster: @zaifastafa 

> @TheKeymaster OK it makes sense now. $criteria->hasCondition($criteriaFacet->getName()) So what does this sentence in the description mean?
>
>Pass the facet's name or the facet's field name if it's value is price.

This can be ignored, I think this was overseen by Isi who has planned this story, but when getting the name of the price it always should be price, because we have a constant in there, so it will never change.

> @TheKeymaster also, can you describe a bit what is the logic of / difference between "selected filters", "facets" and "filter items". How are they interconnected?

* "Selected filters" are filters that were chosen to be filtered by the customer. E.g. `Shirt Size=XL`.
* "Facets" are filters in Shopware. E.g. `Shoe Size` or `vendor`.
* "Filter items" are filter values directly from the XML response. This is typically a list of all available filter values for a certain filter. E.g. vendor has filter items such as: "Nike", "Puma", etc. Furthermore once a filter is selected and the filter is not multiselect, this filter may be no longer returned from FINDOLOGIC (XML_2.0 response).

zaifastafa: @TheKeymaster OK, so for e.g in this scenarios: 

> Filters which are selected only contain the value that was selected

we will NOT have the filter items returned in the XML response and instead we will check the filter from the condition? 

OR is it the opposite in a case such as: 

> All filter values are displayed if no filters are selected

TheKeymaster: @zaifastafa 

> > Filters which are selected only contain the value that was selected

> we will NOT have the filter items returned in the XML response and instead we will check the filter from the condition?

The FINDOLOGIC response will not return the selected filter. Instead we check if Shopware knows if this Facet is selected. If it is, we know that this filter is indeed selected.

> OR is it the opposite in a case such as:
> > All filter values are displayed if no filters are selected

All filter values that were returned by FINDOLOGIC should be displayed if no filter value is selected or the filter is multiselect (in which case you would be able to select multiple values at once).

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -161,17 +165,32 @@ protected function createFacets(Criteria $criteria, SimpleXMLElement $filters)
...
+        /** @var ProductAttributeFacet $criteriaFacet */
         foreach ($criteria->getFacets() as $criteriaFacet) {
             $field = $criteriaFacet->getField();
-            $filter = $filters->xpath(sprintf('//name[.="%s"]/parent::*', $field));
-            if (empty($filter)) {
-                continue;
+            $selectedFilter = $filters->xpath(sprintf('//name[.="%s"]/parent::*', $field));
+
+            if (empty($selectedFilter)) {
+                if ($criteria->hasUserCondition($criteriaFacet->getName())) {
```
TheKeymaster: Debugging this must be hell. There are so many conditions and a dangling else. Consider some defensive coding e.g. (variable and function names may be changed)
```php
protected function createFacets(Criteria $criteria, SimpleXMLElement $filters)
{
    // . . . foreach . . .
    $selectedFilter = $selectedFilterByResponse = $this->fetchSelectedFilterByResponse($filters, $field);
    if (!$selectedFilterByResponse) {
        $selectedFilter = $this->fetchSelectedFilterByUserCondition($criteriaFacet);
        if (!$selectedFilter) {
            continue;
        }
    }
    // . . .
}

private function fetchSelectedFilterByResponse(SimpleXMLElement $filters, $field)
{
    $selectedFilter = $filters->xpath(sprintf('//name[.="%s"]/parent::*', $field));
    return !empty($selectedFilter) ? $selectedFilter[0] : null;
}

private function fetchSelectedFilterByUserCondition(Criteria $criteria, FacetInterface $criteraFacet)
{
    if (!$criteria->hasUserCondition($criteriaFacet->getName())) {
        return null;
    }

    $condition = $criteria->getUserCondition($criteriaFacet->getName());
    return $this->createSelectedFilter(
        $criteriaFacet,
        $condition
    );
}
```

FinSearchUnified/Bundle/SearchBundleFindologic/FacetHandler/TextFacetHandler.php
```diff
@@ -122,11 +122,9 @@ private function getValueListItems(FacetInterface $facet, Criteria $criteria, Si
      */
     private function createValueListFacetResult(FacetInterface $facet, Criteria $criteria, SimpleXMLElement $filter)
     {
-        if (!$filter->items->item) {
-            return null;
-        }
+        $items = $filter->items->item;
```
TheKeymaster: Why was this variable introduced? It is not consistent with `createRadioFacetResult`.

zaifastafa: @TheKeymaster you told me in a previous review to introduce this variable which checks for the null value

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -161,17 +165,32 @@ protected function createFacets(Criteria $criteria, SimpleXMLElement $filters)
...
+                    $condition = $criteria->getUserCondition($criteriaFacet->getName());
+                } elseif ($criteria->hasUserCondition('price')) {
```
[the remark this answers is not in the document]
zaifastafa: @TheKeymaster Based on the description, we need to pass a `PriceCondition`, which will never have a name similar to the `ProductAttributeFacet` such as `product_attribute_price` so the 'price' condition is never tested in this scenario. Am I missing something?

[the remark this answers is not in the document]
zaifastafa: I had tried without this separate condition but our PriceCondition was never tested due to the name being different as we are checking the condition from the facetName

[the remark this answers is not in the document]
TheKeymaster: @zaifastafa but the name of the price condition is always `price` and that should work without any issues. I will add a video in which I have removed this condition and it seems to work.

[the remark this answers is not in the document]
zaifastafa: @TheKeymaster yes the name is price if we use the `PriceCondition`, but the `hydrateFacet` generates the `ProductAttributeFacet` which has the name as `product_attribute_price` whereas the condition of the price has the name as `price`

PR #273, opened 2019-11-08
SW-349 Test plugin for lowest compatibility version v5.2.0

zaifastafa:


TheKeymaster:
Click-Tested and works with:

* Shopware v5.2.10.

Earlier versions like `v5.2.0` do not really work with the `shopware-docker` repository.

PR #276, opened 2019-12-04
Sw 358 configuring empty attribute values must not break the export

zaifastafa:


FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -54,7 +54,7 @@ private function createTestProduct(array $testProductConfiguration)
 
             return $article;
         } catch (Exception $e) {
```
TheKeymaster: Can you remember why we are catching this exception here? What is the reason we are even echo'ing it out?

Will the tests break if we removed this catch? Please find out what purpose it has and why this is needed.
We can be better than this.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -90,12 +90,12 @@ public function articleSupplierProvider()
...
                 ]
             ],
-            'Article with supplier name space' => [
+            'Article with pseudo empty supplier name' => [
                 [
                     'name' => 'FindologicArticle 2',
                     'active' => true,
                     'tax' => 19,
-                    'supplier' => " ",
+                    'supplier' => ' ',
```
TheKeymaster: Can we also check multiple spaces?

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
zaifastafa: @TheKeymaster Hi, I dont think I was not the one who implemented this case, but from what I can tell it is just for debugging to let us know if the test article was not created due to some missing or invalid configuration which we are passing. 
If the article is not created then it would eitherways break the test as the article is needed to be created which is why the method is called. But it would not fail purely due to the missing exception.

PR #278, opened 2019-12-06
Sw 286 smart suggest result should consider inactive filters

zaifastafa:


FinSearchUnified/Bundle/StoreFrontBundle/Gateway/Findologic/CustomFacetGateway.php
```diff
@@ -120,13 +122,33 @@ public function getAllCategoryFacets(ShopContextInterface $context)
...
+                $hasVendorFacet = true;
+            }
+            if ($facetName === 'cat') {
+                $hasCategoryFacet = true;
+            }
+
+            $facets[] = $facet;
+        }
+
+        if ($hasCategoryFacet === false) {
```
TheKeymaster: Checking for falseness is enough.

FinSearchUnified/Bundle/StoreFrontBundle/Gateway/Findologic/CustomFacetGateway.php
```diff
@@ -120,13 +122,33 @@ public function getAllCategoryFacets(ShopContextInterface $context)
...
+            $facets[] = $this->hydrator->hydrateDefaultCategoryFacet();
+        }
+        if ($hasVendorFacet === false) {
```
TheKeymaster: Checking for falseness is enough.

FinSearchUnified/Tests/Components/ConfigLoaderTest.php
```diff
@@ -173,18 +195,18 @@ public function testWarmUpCache($responseCode, $responseBody, array $expectedCon
...
-            'Config file is empty or null' => [200, null, []],
-            'Config file returns extra parameters' => [
-                200,
-                json_encode([
-                    'isStagingShop' => true,
-                    'directIntegration' => ['enabled' => false, 'iShouldNotBeHere' => true]
-                ]),
-                ['isStagingShop' => true, 'directIntegration' => ['enabled' => false]]
-            ],
+            'Cache load return false' => [false, $this->exactly(2)],
```
TheKeymaster: Remove this data provider and split the test up in two cases. These are two separate scenarios, they are checking a different result so they should be separate tests.

FinSearchUnified/Tests/Bundle/StoreFrontBundle/Gateway/Findologic/Hydrator/CustomListingHydratorTest.php
```diff
@@ -279,4 +284,94 @@ public function testFilterNamesAreEscaped($text, $expected, $errorMessage)
...
         $this->assertEquals($expected, $result, $errorMessage);
     }
+
+    /**
+     * @throws ReflectionException
+     */
+    public function testCreateCustomFacet()
+    {
+        $name = 'Price';
+        $label = 'Pries';
```
TheKeymaster: Looks like a typo.

FinSearchUnified/Tests/Components/ConfigLoaderTest.php
zaifastafa: I am sorru but how is this a separate scenario? We are checking for the outcome of the cache load method, where the first scenario does not return the cache, whereas the second does return a cache.

FinSearchUnified/Bundle/StoreFrontBundle/Gateway/Findologic/CustomFacetGateway.php
zaifastafa: Were you able to identify why the default facet method needs to be called twice instead of once? I had mentioned this on youtrack about the issue. From what i think it should pass with an expectation of once but it is passing for exactly(2) instead.

FinSearchUnified/Tests/Components/ConfigLoaderTest.php
TheKeymaster: Yes so you basically have two different results. In the first scenario you want to ensure that `save` is called and in the other one you want to ensure that `save` is never called. Having an if inside of a test already tells you that it's probably a better idea to split it up.

A test should always test the same scenario and shouldn't do things based on what data the dataprovider returns. 

For example:

```php
public function someProvider()
{
    return array(
        array(0, 0, 0),
        array(0, 1, 1),
        array(1, 0, 1),
        array(1, 1, 3)
    );
}
```

Bad test case:

```php
/**
 * @dataProvider someProvider
 */
public function testAdd($a, $b, $c)
{
    if ($a = 0) {
        $this->assertNotEquals($c, $a + $b);
    } else {
        $this->assertEquals($c, $a + $b);
    }
}
```

Good test case:

```php
/**
 * @dataProvider someProvider
 */
public function testAdd($a, $b, $c)
{
    $this->assertEquals($c, $a + $b);
}
```

Of course this is an oversimplified example, but please split up the test if your assertions differ.

zaifastafa: @TheKeymaster OK I guess I understand the cases. But then this would require a lot of refactoring in other test cases as well because a lot of tests are written in this way. So maybe we can do this optimization and refactoring in a future story if possible.

PR #282, opened 2019-12-27
SW-347-export-products-without-category

zaifastafa:


TheKeymaster:
Closing this PR as Shopware default does not support this anyway. These exported products wouldn't show up in a search result.

PR #283, opened 2020-01-01
SW-344-redirect-shopware-search-api-failure

zaifastafa:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -68,33 +68,33 @@ public function search(Criteria $criteria, ShopContextInterface $context)
         $response = $query->execute();
 
         if (empty($response)) {
-            self::setFallbackFlag(1);
+            self::setFallbackSearchFlag(1);
```
TheKeymaster: In the implementation I wrote "Instead of the old implementation do", but please still set the fallback flag to `1` (like before).

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
```diff
@@ -90,47 +136,58 @@ public function testProductNumberSearchImplementation($isFetchCount, $isUseShopS
...
 
         $context = Shopware()->Container()->get('shopware_storefront.context_service')->getContext();
         $productNumberSearch->search($criteria, $context);
     }
 
-    public function productNumberSearchProvider()
+    /**
+     * @throws Enlight_Exception
+     */
+    public function testFallBackCookieIsSet()
```
TheKeymaster: True, we can not get around the redirection. Let's keep it still. Maybe we know more in the future.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -251,8 +258,39 @@ public static function shopSearchProvider()
                 'findologicDI' => false,
                 'isSearchPage' => false,
                 'isCategoryPage' => true,
+                'fallbackSearchCookie' => null,
                 'expected' => false
-            ]
+            ],
+            'Cookie "fallback-search" is set and true' => [
```
TheKeymaster: Add another test case, in which the fallbackSearchCookie is `1`.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -911,4 +914,12 @@ public static function setSmartDidYouMean(SimpleXMLElement $xmlResponse)
             );
         }
     }
+
+    /**
+     * @return bool
+     */
+    private static function checkIfFallbackSearchCookieIsSet()
+    {
+        return (isset($_COOKIE['fallback-search']) && $_COOKIE['fallback-search']) === true;
```
TheKeymaster: Cast `$_COOKIE['fallback-search'])` to a `bool` before checking for `=== true`.

PR #286, opened 2020-01-20
SW-375-fix-warnings-shopware-code-review

zaifastafa:


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -786,11 +785,11 @@ protected function setProperties()
     }
 
     /**
-     * @param ArticleAttribute $attributes
+     * @param $attributes
```
TheKeymaster: * Was the type before wrong?
* What would be the correct type?
* Why wasn't the correct type added?

zaifastafa: @TheKeymaster 
The `Attribute` models are actually not defined in the Shopware namespace so it was giving a warning. The namespace is somehow generated automatically so you would not find an actual class under that namespace. It is a bit confusing and I do not know why this happens. But if they are giving it as an error, I thought removing it would fix it as it does not affect the functionality in anyway.

TheKeymaster: Hm... okay I see, that's unfortunate. I will ask Shopware what their suggestion on this is. For now let's leave it as it is.

PR #287, opened 2020-01-20
SW-376-add-config-field-cross-sell-categories

zaifastafa:


FinSearchUnified/Resources/config.xml
```diff
@@ -27,6 +27,18 @@
             <description lang="de">Aktiviert die FINDOLOGIC Suche für Kategorieseiten.</description>
             <description lang="en">Activate the FINDOLOGIC search provider for category pages.</description>
         </element>
+        <element type="combo">
+            <name>CrossSellingCategories</name>
```
TheKeymaster: Please ensure that this works with all Shopware versions that we support. Shopware 5.2 should work as well.

FinSearchUnified/Resources/config.xml
```diff
@@ -27,6 +27,18 @@
             <description lang="de">Aktiviert die FINDOLOGIC Suche für Kategorieseiten.</description>
             <description lang="en">Activate the FINDOLOGIC search provider for category pages.</description>
         </element>
+        <element type="combo">
```
TheKeymaster: @zaifastafa I click-tested this on my own with Shopware 5.6 and it seems to work fine. On Shopware 5.2 it also works, but I am not able to select sub-categories.

image

The moment I type anything that does not match any of those two categories (e.g. a sub category), it does not work.
Please ensure that you can also select sub-categories.

FinSearchUnified/Resources/config.xml
```diff
@@ -27,6 +27,35 @@
...
+    proxy: {
+        type: 'ajax',
+        url: window.location.href.substr(0, window.location.href.indexOf('backend')) + 'backend/Category/getPathByQuery',
+        reader: {
+            type: 'json',
+            root: 'data',
+            totalProperty: 'total'
+        }
+    }
+    }).create();//new ]]>
```
TheKeymaster: I do not think that this comment `//new` is necessary.

zaifastafa: It is mentioned in their documentation that it is a sort of hack to load the store.

TheKeymaster: I see. If it works, then ok :).

PR #291, opened 2020-01-29
SW-385 Make Product Stream preview great again

TobiasGraml11:


FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -204,7 +204,7 @@ public static function useShopSearch()
     {
         $request = Shopware()->Front()->Request();
         $isCLIMode = $request === null;
-        if ($isCLIMode) {
+        if ($isCLIMode || !Shopware()->Container()->has('shop')) {
```
TheKeymaster: Add a comment above the condition, mentioning when no shop is available, and why we need this check.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -204,7 +204,12 @@ public static function useShopSearch()
     {
         $request = Shopware()->Front()->Request();
         $isCLIMode = $request === null;
-        if ($isCLIMode) {
+
+        /*
+         * Shop is not available in the Backend for the search of the Product Stream preview.
```
TheKeymaster: Sorry that I have to add a review issue for that, but we decided a long time ago on how comments should be written.
`//` is the way to go. E.g.

```php
// Bla bla bla, this is really important information,
// removing the following line will destroy the whole universe!
```

PR #292, opened 2020-02-05
Sw 384 filters should be updated page reloaded

zaifastafa:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -163,13 +174,28 @@ private function getFacetHandler(SimpleXMLElement $filter)
...
+        if ($this->cache->load($cacheId) === false) {
+            /** @var QueryBuilder $query */
+            $query = $this->queryBuilderFactory->createSearchNavigationQuery($criteria, $context);
+            $response = $query->execute();
+            $this->cache->save($response, $cacheId, ['FINDOLOGIC'], 86400);
+        } else {
+            $response = $this->cache->load($cacheId);
+        }
+
+        $xmlResponse = StaticHelper::getXmlFromResponse($response);
```
TheKeymaster: Create a separate method for this code block.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -388,4 +388,14 @@ private static function checkIfFallbackSearchCookieIsSet()
     {
         return (isset($_COOKIE['fallback-search']) && (bool)$_COOKIE['fallback-search']) === true;
     }
+
+    /**
+     * @return bool
+     */
+    public static function checkIfProductAndFilterLiveReloadingIsEnabled()
```
TheKeymaster: `isProductAndFilterLiveReloadingEnabled`

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -363,7 +363,7 @@ public static function setSmartDidYouMean(SimpleXMLElement $xmlResponse)
         $originalQuery = (string)$query->originalQuery;
         $didYouMeanQuery = (string)$query->didYouMeanQuery;
         $queryString = $query->queryString;
-        $queryStringType = (string)$queryString->attributes()->type;
+        $queryStringType = $queryString->attributes() !== null ? (string)$queryString->attributes()->type : null;
```
TheKeymaster: Why was this changed? It has nothing to do with this story.

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
```diff
@@ -9,25 +9,34 @@
...
+use Shopware\Bundle\SearchBundleDBAL\ProductNumberSearch as OriginalProductNumberSearch;
+use Shopware\Bundle\StoreFrontBundle\Service\ContextServiceInterface;
 use Shopware_Components_Config as Config;
 use SimpleXMLElement;
+use Zend_Cache_Core;
 
 class ProductNumberSearchTest extends TestCase
 {
+    /**
+     * @var \Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface
```
TheKeymaster: This can be imported.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -630,7 +630,7 @@ public function testBuildCategoryName($categoryId, $category, $expected)
     private function updateParentCategoryName(Category $parent, $restore = true)
     {
         // Stop when Shopware's root category is reached. Changing it can and will break unrelated tests.
-        if ($parent->getId() === 1) {
+        if ($parent->getId() <= 3) {
```
TheKeymaster: Why was this change necessary?

FinSearchUnified/composer.lock
```diff
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "findologic/libflexport",
-            "version": "v1.2.2",
+            "version": "v1.4.0",
```
TheKeymaster: Why was the version updated? Please revert this change.

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -163,13 +174,28 @@ private function getFacetHandler(SimpleXMLElement $filter)
...
     {
         $facets = [];
+        $url = md5(Shopware()->Front()->Request()->getRequestUri());
+        $cacheId = sprintf('finsearch_%s', $url);
+
+        if ($this->cache->load($cacheId) === false) {
+            /** @var QueryBuilder $query */
+            $query = $this->queryBuilderFactory->createSearchNavigationQuery($criteria, $context);
+            $response = $query->execute();
+            $this->cache->save($response, $cacheId, ['FINDOLOGIC'], 86400);
```
TheKeymaster: Nobody knows what `86400` really is. It is much easier by just putting a formula like `60 * 60 * 24`, because everyone knows that this is a day, by just looking at it.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
zaifastafa: @TheKeymaster I was trying to figure out something as whenever I execute the tests, I get like loads of shop entries created in the s_core_shops table and some "Deutsch" categories in the s_categories table. This over the time makes the tests really really slow as there are a lot of shop entries it has to go through. 
Can you please verify it at your end as well. I had discussed this with Ismail as well but he could not figure out it either why this was happening. Maybe you would be able to identify something related to it.

FinSearchUnified/Helper/StaticHelper.php
zaifastafa: This was causing the tests to fail as I introduced the demoResponse.xml and it does not have attributes in it. So this was just a precautionary error handling.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
TheKeymaster: @zaifastafa when are these entries created? Are they created when you execute a specific test case? Please investigate this further.

FinSearchUnified/Helper/StaticHelper.php
TheKeymaster: Changing production code, to make tests work is generally speaking a bad idea. There are only two cases where this change would've been necessary.

1. The response is valid and the code was never working at all. => Adapt the code, so this case is handled properly.
2. The response is invalid and the code threw an exception, because he didn't expect a malformed response. => Adapt the response.

Since nobody has ever complained about this, nor anyone ever had this error, I am pretty sure that the response is invalid or malformed.

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilderFactory.php
```diff
@@ -250,4 +250,33 @@ public function createQueryBuilder()
...
     }
+
+    /**
+     * @param Criteria $criteria
+     * @param ShopContextInterface $context
+     *
+     * @return QueryBuilder
+     * @throws Exception
+     */
+    public function createSearchNavigationQuery(Criteria $criteria, ShopContextInterface $context)
```
TheKeymaster: Explicitly mention that no additional filters are sent, maybe even in the method name itself.

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -298,4 +342,28 @@ protected static function redirectToSameUrl()
...
     }
+
+    /**
+     * @param Criteria $criteria
+     * @param ShopContextInterface $context
+     *
+     * @return false|mixed|string|null
+     * @throws Zend_Cache_Exception
+     */
+    private function getResponse(Criteria $criteria, ShopContextInterface $context)
```
TheKeymaster: `getResponse` is way to general. Rename this to `getResponseWithoutFilters`.

TheKeymaster:
@zaifastafa please also fix your merge conflicts.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
zaifastafa: @TheKeymaster am not able to figure out which test causes this problem. I know its some test in StaticHelperTest, and it only happens when all tests are executed together. But its not guaranteed that these entries are created everytime. It does not seem to have any pattern AFAIK. Sometimes it creates these entries, sometimes it doesn't which makes it difficult to identify the root. 

I had mentioned this to Ismail as well but he was not able to identify the problem either.

FinSearchUnified/Helper/StaticHelper.php
zaifastafa: @TheKeymaster I am not aware of how the response is generated. I only have the demo response which was provided by @dblaichinger and in that the `attributes` was not available. In another case, we were also checking for the `type` of attribute, so based on the information I have, I updated the code to handle either of cases in the response; whether the attributes is there, or not, our code would then not fail.

TheKeymaster: I just checked the response. The change seems to be fine.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
TheKeymaster: I'd suggest changing it locally when it happens, instead of changing it for everyone, because I also haven't experienced such an issue when running tests. Please revert this change and add it if you need it locally, to improve the performance of your tests.

FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -21,6 +22,7 @@
 use Shopware\Bundle\SearchBundleDBAL\QueryBuilderFactoryInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
 use SimpleXMLElement;
+use Symfony\Component\HttpFoundation\Request;
```
TobiasGraml11: This class is not used

PR #296, opened 2020-02-10
SW-377-exclude-products-with-cross-sell-category

zaifastafa:


zaifastafa:
@TheKeymaster I have made changes to the `testbuildCategoryName` in StaticHelperTest class due to the reason I had mentioned earlier that additional shop and category entries were being created during tests execution. This was due to the changes being done in default shop categories in the test which was affecting other test cases in some scenarios and was not only affected locally but it happened on travis too, which is why removing that logic fixed the failing build.

FinSearchUnified/Tests/Components/ConfigLoaderTest.php
```diff
@@ -244,14 +244,12 @@ public function testCacheWithCorrectKey($cacheTestResponse, $loadCallCount)
     public function testCacheWithNonExistingKey($cacheResponse, $loadCallCount)
     {
         $httpClient = Shopware()->Container()->get('http_client');
```
TheKeymaster: This change should not have been necessary. Please re-add it.

FinSearchUnified/Tests/PluginTest.php
```diff
@@ -158,11 +204,9 @@ public function testEmptyValueNotAllowedExceptionIsThrownInExport()
     private function runExportAndReturnCount()
     {
         try {
-            $shopKey = 'ABCDABCDABCDABCDABCDABCDABCDABCD';
-            /** @var ShopwareProcess $blController */
-            $blController = Shopware()->Container()->get('fin_search_unified.shopware_process');
-            $blController->setShopKey($shopKey);
-            $xmlDocument = $blController->getFindologicXml();
+            $shopwareProcess = Shopware()->Container()->get('fin_search_unified.shopware_process');
```
TheKeymaster: I think the docblock makes sense here. This way the IDE knows the class of the instance. But good idea to rename this variable name though :D.

FinSearchUnified/Tests/PluginTest.php
```diff
@@ -90,13 +89,52 @@ public function testArticleExport($articlesActiveStatus, $expectedCount, $errorM
...
+            ],
+            'Article does not exist in cross-sell category configured' => [
+                'crossSellingCategories' => [5],
+                'expectedCount' => 1
+            ],
+            'Article exists in one of the cross-sell categories configured' => [
+                'crossSellingCategories' => [8, 9],
+                'expectedCount' => 1
+            ],
+            'Article exists in all of cross-sell categories configured' => [
```
TheKeymaster: I think it was a little but confusing, but actually the count should be `0` for all cases, except for the first one. As long as one cross sell category matches the product, it should not be exported.
The product has the categories 8, 9 and 10, which means if any of these categories are declared as a cross-sell category, the product should not be exported.

For example:
This means if the category 5 is assigned, which may be a child of 1 and the configured cross-sell category is 1, then the product should still be exported. If on the other hand, 5 is set as cross-sell category, the product should not be exported.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -600,60 +599,12 @@ public function testCleanStringMethod($text, $expected, $errorMessage)
...
     public function testBuildCategoryName($categoryId, $category, $expected)
     {
-        $categoryModel = $this->categoryResource->update($categoryId, [
-            'name' => $category
-        ]);
-        $this->assertInstanceOf(Category::class, $categoryModel);
-
-        $this->updateParentCategoryName($categoryModel->getParent(), false);
-
+        $categoryModel = $this->categoryResource->update($categoryId, ['name' => $category]);
```
TheKeymaster: Why cant we reset our changes to the database after a test has been completed e.g. `tearDown`?

zaifastafa: It is specific to only this case. We don't want to update the database for every case in this class.

FinSearchUnified/Tests/Components/ConfigLoaderTest.php
zaifastafa: It was necessary because the shopkey is changed mid-execution which can cause the expectation defined in this test to fail - since the cacheKey is dependent on the shopkey then it changes the generated key which can cause problems. 
This change does not change the outcome as it is already being tested.

FinSearchUnified/Tests/PluginTest.php
zaifastafa: @TheKeymaster https://findologic.myjetbrains.com/youtrack/issue/SW-377#focus=streamItem-94-24651.0-0
This would require re-implementing the story as I dont think it would be possible through query alone. We will have to use the `shouldBeExported` flag for this which makes the current implementation useless.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -173,6 +173,8 @@ public function __construct(
             $this->setVariantOrdernumbers();
             $this->setProperties();
         }
+
+        $this->isCrossSellingCategoryConfigured();
```
TheKeymaster: This should done before setting all fields.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -837,4 +839,16 @@ public function getXmlRepresentation()
     {
         return $this->xmlArticle;
     }
+
+    private function isCrossSellingCategoryConfigured()
```
TheKeymaster: Looking at this method makes me think it returns a boolean. Make this return a boolean and set `$this->shouldBeExported` to the return value of this method. Stop the execution in the constructor by `return`ing if this method returns `true` (product has a cross sell category).

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
zaifastafa: It will just get overridden in the `setVariant` method.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
zaifastafa: OK, in that case ignore my previous comment.

PR #301, opened 2020-02-18
Sw 374 clicking inactive filter response should not skip

zaifastafa:


TheKeymaster:
@zaifastafa I am having quite a hard time going through this PR without missing any issues. Could you merge `develop` into `SW-286` and then merge `SW-286` in this one? This way I can see the changes much clearer and can focus on the real changes instead of going through 3000 added and 1500 deleted lines.

zaifastafa:
> @zaifastafa I am having quite a hard time going through this PR without missing any issues. Could you merge `develop` into `SW-286` and then merge `SW-286` in this one? This way I can see the changes much clearer and can focus on the real changes instead of going through 3000 added and 1500 deleted lines.

@TheKeymaster OK sure. So the base branch for the current SW-374 branch would be SW-286 or develop then?

PR #302, opened 2020-02-19
SW-367-show-suitable-search-result-title

zaifastafa:


FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -392,4 +392,80 @@ private static function checkIfFallbackSearchCookieIsSet()
...
+            $snippetType = 'cat';
+        } elseif (isset($params['vendor']) && !empty($params['vendor'])) {
+            $vendor = $params['vendor'];
+            $filterName = $snippetManager->get('frontend/search/query_info_message/filter_vendor');
+            $snippetType = 'vendor';
+        } else {
+            $snippetType = 'default';
+        }
+
+        return [$vendor, $cat, $smartQuery, $filterName, $snippetType];
```
TheKeymaster: I really do not like the idea of arrays that need to be ordered in a certain way without any structure that holds them together. This looks like an avoidance of OOP, since this structure could be moved to a class. That being said, I also think that the logic from above would make more sense to be in a class.

FinSearchUnified/Resources/views/frontend/search/fuzzy.tpl
```diff
@@ -2,15 +2,16 @@
 
 {block name="frontend_search_headline"}
     <h1 class="search--headline">
-        {if $finSmartDidYouMean == false || $finSmartDidYouMean.type == 'did-you-mean'}
-            {s name='SearchHeadline'}{/s}
+        {if $snippetType == 'query'}
+            {s namespace='frontend/search/query_info_message' name='frontend/search/query_info_message/query'}{/s}
```
TheKeymaster: Like we do for smart did you mean, add the English translation in between of `{s}{/s}`. This way the translations work even if the snippets do not yet exist, in case the plugin hasn't been reinstalled.

FinSearchUnified/Helper/SnippetExtractor.php
```diff
@@ -0,0 +1,165 @@
+<?php
+
+namespace FinSearchUnified\Helper;
+
+use Enlight_View_Default;
+use SimpleXMLElement;
+
+class SnippetExtractor
```
TheKeymaster: This is more like a `QueryInfoMessageParser` than an extractor. But nice restructuring ;)!

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -392,4 +392,27 @@ private static function checkIfFallbackSearchCookieIsSet()
...
+
+    /**
+     * @param SimpleXMLElement $xmlResponse
+     */
+    public static function setQueryInfoMessage(SimpleXMLElement $xmlResponse)
+    {
+        /** @var Enlight_View_Default $view */
+        $view = Shopware()->Container()->get('front')->Plugins()->get('ViewRenderer')->Action()->View();
+
+        $snippetExtractor = new SnippetExtractor($xmlResponse, $view);
```
TheKeymaster: Also rename this variable after renaming the class.

PR #306, opened 2020-03-24
SW-408 A no result page should not throw an exception

TobiasGraml11:


FinSearchUnified/Bundle/ProductNumberSearch.php
```diff
@@ -286,10 +286,14 @@ private function fetchSelectedFilterByUserCondition(
 
         $condition = $criteria->getUserCondition($facetName);
 
-        return $this->createSelectedFilter(
-            $criteriaFacet,
-            $condition
-        );
+        if ($condition) {
```
TheKeymaster: Try to apply [Defensive Coding](https://en.wikipedia.org/wiki/Defensive_programming) in this case. I'd suggest returning null only if there is no condition.

```php
if (!$condition) {
    return null;
}

// Do the usual stuff.
```

TheKeymaster: Maybe it makes sense to point out in a comment, when a condition may not be able to be fetched.

TobiasGraml11: Isn't it clear, when the line above is trying to get a `UserCondition`?

TheKeymaster: That's why I said "maybe". Judge yourself if it is necessary :).

PR #308, opened 2020-03-26
SW-405-communicate-with-findologic-api

zaifastafa:
The build is failing as the findologic-api PR is not merged yet. It will pass once the PR is merged into ~1.2

### DIFF
--- a/FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder/NewQueryBuilder.php
+++ b/FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder/NewQueryBuilder.php
@@ -170,8 +170,8 @@ abstract class NewQueryBuilder
      */
     public function addUserGroup($usergroup)
     {
-        $hashedGroup = StaticHelper::calculateUsergroupHash($this->shopKey, $usergroup);
-        $this->searchNavigationRequest->addGroup($hashedGroup);
+        $usergrouphash = StaticHelper::calculateUsergroupHash($this->shopKey, $usergroup);
+        $this->searchNavigationRequest->addUserGroup($usergrouphash);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder/NewQueryBuilder.php
```diff
@@ -0,0 +1,240 @@
...
+        $this->searchNavigationRequest->addAttribute($key, $value);
+    }
+
+    /**
+     * @param string $usergroup
+     */
+    public function addUserGroup($usergroup)
+    {
+        $hashedGroup = StaticHelper::calculateUsergroupHash($this->shopKey, $usergroup);
+        $this->searchNavigationRequest->addGroup($hashedGroup);
```
TheKeymaster: This one should actually be `addUserGroup`. I think FINDOLOGIC-API does not yet support this parameter, so please add another PR that adds this parameter. Thanks.

TheKeymaster: The name of the parameter is `usergrouphash` and the method should be `addUserGroup`.

FinSearchUnified/composer.json
```diff
@@ -21,7 +21,8 @@
   },
   "require": {
     "php": ">=5.6.4",
-    "findologic/libflexport": "~1.2"
+    "findologic/libflexport": "~1.2",
+    "findologic/findologic-api": "~1.2"
```
TheKeymaster: You should be able to get `1.3` now.

FinSearchUnified/Subscriber/RegisterComponents.php
```diff
@@ -2,6 +2,7 @@
 
 namespace FinSearchUnified\Subscriber;
 
+use Composer\Autoload\ClassLoader;
```
TheKeymaster: This import is not needed.

zaifastafa: @TheKeymaster why so?

PR #310, opened 2020-04-02
SW-402 Make Crossselling Categories compatible

TobiasGraml11:
In older versions of Shopware, the category name is saved instead of the ID.

Because of that, we need to check for the category name instead


dblaichinger:
@TheKeymaster Can you please create a story for this and add the id to the PR. Thanks!

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -864,9 +864,19 @@ protected function isCrossSellingCategoryConfiguredForArticle()
...
         foreach ($this->baseArticle->getCategories() as $category) {
-            if (in_array($category->getId(), $crossSellingCategories, true)) {
+            if (in_array($this->buildCategoryTree($category), $crossSellingCategories, true)) {
                 return true;
             }
         }
+        return false;
+    }
+
+    protected function buildCategoryTree($category)
```
TheKeymaster, line 874: Add a typehint to `$category` (not a docblock).

PR #316, opened 2020-05-04
SW-415-staging-button

zaifastafa:


FinSearchUnified/Controllers/Backend/FindologicStaging.php
```diff
@@ -0,0 +1,75 @@
...
+
+            Shopware()->Plugins()->Controller()->ViewRenderer()->setNoRender(true);
+            $shopkey = Shopware()->Config()->offsetGet('ShopKey');
+            $isActive = Shopware()->Config()->offsetGet('ActivateFindologic');
+
+            if (!$isActive) {
+                throw new RuntimeException('Please ensure the plugin is active!');
+            }
+            if (!preg_match('/^[A-F0-9]{32}$/', $shopkey)) {
+                throw new RuntimeException('Please ensure a valid ShopKey is provided!');
```
TheKeymaster, line 26: You have not imported the `RuntimeException` class.

FinSearchUnified/Resources/config.xml
```diff
@@ -83,6 +83,13 @@
                 <readOnly>true</readOnly>
             </options>
         </element>
+        <element type="button" scope="shop">
```
TheKeymaster, line 86: Please check how we can make this work for Shopware < 5.2.17 If you can not figure it out or it just does not work, you should still be able to change/install the Plugin when using Shopware < 5.2.17, therefore the button may be removed completely in that case.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
```diff
@@ -0,0 +1,75 @@
...
+            $configLoader = Shopware()->Container()->get('fin_search_unified.config_loader');
+            try {
+                $isStaging = $configLoader->isStagingShop();
+            } catch (\Exception $e) {
+                $isStaging = false;
+            }
+            if ($isStaging) {
+                $queryParam = '?findologic=on';
+            }
+            $this->redirect(
```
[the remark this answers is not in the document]
zaifastafa: @TheKeymaster yes have already checked, it does open a new tab.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
```diff
@@ -0,0 +1,75 @@
...
+                sprintf(
+                    '%s://%s%s',
+                    $shop->getSecure() ? 'https' : 'http',
+                    rtrim($shop->getHost(), '/'),
+                    $queryParam
+                )
+            );
+        } catch (Exception $e) {
+            echo $e->getMessage();
```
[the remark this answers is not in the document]
zaifastafa: @TheKeymaster How do you want to handle the exception in this scenario? Should I just let it throw an exception on opening the link?

FinSearchUnified/Controllers/Backend/FindologicStaging.php
```diff
@@ -0,0 +1,75 @@
...
+            }
+            $this->redirect(
+                sprintf(
+                    '%s://%s%s',
+                    $shop->getSecure() ? 'https' : 'http',
+                    rtrim($shop->getHost(), '/'),
+                    $queryParam
+                )
+            );
+        } catch (Exception $e) {
```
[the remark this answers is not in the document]
zaifastafa: It does not need to be imported. It is from the default namespace.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
zaifastafa: It does not need to be imported. It is from the default namespace.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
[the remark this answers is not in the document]
TheKeymaster: Yes Shopware will handle it anyway somehow. We shouldn't worry about that. It is not our responsibility to catch all Exceptions that may occur at any point of execution. It is our responsibility to catch all implementation-specific Exceptions that may occur e.g. wrong user input.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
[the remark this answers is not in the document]
TheKeymaster: It is bad practice to not import a namespace from the global namespace. This code alone could cause issues if someone was to create the class `Exception` in `FinSearchUnified/Controllers/Backend`. In that case the exception from the current namespace would be used, which is not clearly visible and should not happen at all.

Either import classes from the global namespace or prefix them with `\`.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
TheKeymaster: See my other comment.

FinSearchUnified/FinSearchUnified.php
```diff
@@ -54,6 +56,29 @@ public function install(InstallContext $context)
...
+            $file = $this->getPath() . '/Resources/config/config_compat.xml';
+        }
+
+        if (class_exists(XmlConfigReader::class)) {
+            $xmlConfigReader = new XmlConfigReader();
+        } else {
+            $xmlConfigReader = new \Shopware\Components\Plugin\XmlConfigDefinitionReader();
+        }
+
+        $config = $xmlConfigReader->read($file);
```
TheKeymaster, line 75: Move this into a separate function e.g. `loadConfig`.

FinSearchUnified/FinSearchUnified.php
```diff
@@ -54,6 +56,29 @@ public function install(InstallContext $context)
...
+            $file = $this->getPath() . '/Resources/config/config.xml';
+        } else {
+            // For Shopware <= 5.2.17 we load a different config file due to missing `button` element in configuration
+            $file = $this->getPath() . '/Resources/config/config_compat.xml';
+        }
+
+        if (class_exists(XmlConfigReader::class)) {
+            $xmlConfigReader = new XmlConfigReader();
+        } else {
+            $xmlConfigReader = new \Shopware\Components\Plugin\XmlConfigDefinitionReader();
```
TheKeymaster, line 72: Import this class.

FinSearchUnified/FinSearchUnified.php
```diff
@@ -54,6 +56,29 @@ public function install(InstallContext $context)
     {
         $context->scheduleClearCache([InstallContext::CACHE_TAG_THEME]);
         parent::install($context);
+
+        $shopwareVersion = Shopware()->Config()->get('version');
+
+        if (version_compare($shopwareVersion, '5.2.17', '>=')) {
+            $file = $this->getPath() . '/Resources/config/config.xml';
```
TheKeymaster, line 63: `/Resources/config` could be a constant `CONFIG_DIR`.

FinSearchUnified/FinSearchUnified.php
```diff
@@ -54,6 +56,29 @@ public function install(InstallContext $context)
...
+        } else {
+            // For Shopware <= 5.2.17 we load a different config file due to missing `button` element in configuration
+            $file = $this->getPath() . '/Resources/config/config_compat.xml';
```
TheKeymaster, line 66: Why not just load one config and add the button via code? All options except the button are duplicates.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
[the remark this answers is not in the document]
zaifastafa: I see, alright. The import statement was showing as "unused", but yes I will prefix it with the `\` instead.

FinSearchUnified/FinSearchUnified.php
zaifastafa: Importing this class was trying to find the class from the current namespace, which was failing the test. So I then used FQN instead.

.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
 vendor
+FinSearchUnified/Resources/config.xml
```
TheKeymaster, line 3: Please add a comment in regards of why this is ignored.

FinSearchUnified/FinSearchUnified.php
```diff
@@ -56,6 +63,15 @@ public function install(InstallContext $context)
     {
         $context->scheduleClearCache([InstallContext::CACHE_TAG_THEME]);
         parent::install($context);
+
+        // Manually load the config xml due to Shopware 5.2.0 compatibility
+        $config = $this->loadConfig();
```
TheKeymaster, line 68: Also do this on update.

TheKeymaster: Probably makes sens to put this into an own method. E.g. `initializeConfiguration` or something like that.

FinSearchUnified/Controllers/Backend/FindologicStaging.php
```diff
@@ -0,0 +1,117 @@
...
+
+        $this->redirect($this->getRedirectUrl($isStaging, $shop));
+    }
+
+    private function assertPluginIsActive()
+    {
+        $isActive = Shopware()->Config()->offsetGet('ActivateFindologic');
+
+        if (!$isActive) {
+            throw new \RuntimeException('Please ensure the plugin is active!');
```
TheKeymaster, line 40: Create own Exception classes for that and throw them here. If these exceptions are thrown, the message should be displayed in the frontend. e.g.

```php
throw new StagingModeException('Please ensure that the FinSearchUnified plugin is active.');
```

in the indexAction:

```php
try {

} catch (StagingModeException $e) {
    echo $e->getMessage();
}
```

FinSearchUnified/FinSearchUnified.php
```diff
@@ -102,4 +114,51 @@ private function camelCaseToUnderscore($string)
...
+        $shopwareVersion = Shopware()->Config()->get('version');
+
+        // For Shopware >= 5.2.17 we manually generate the `button` element in configuration as it is not available
+        // for older versions.
+        if (version_compare($shopwareVersion, '5.2.17', '>=')) {
+            $buttonEl = $xml->elements->addChild('element');
+            $buttonEl->addAttribute('type', 'button');
+            $buttonEl->addAttribute('scope', 'shop');
+            $buttonEl->addChild('name', 'StagingTestButton');
+            $buttonEl->addChild('label', 'FINDOLOGIC Test');
```
TheKeymaster, line 135: `FINDOLOGIC Test` => `Findologic Test`

PR #319, opened 2020-05-13
SW-436 Fix check for cross selling categories

TobiasGraml11:


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -864,6 +864,10 @@ protected function isCrossSellingCategoryConfiguredForArticle()
         $crossSellingCategories = Shopware()->Config()->offsetGet('CrossSellingCategories');
         /** @var Category $category */
         foreach ($this->baseArticle->getCategories() as $category) {
+            if (!$category->isChildOf($this->baseCategory)) {
```
TheKeymaster, line 867: Why are we checking that it is not a child of the baseCategory? Doesn't it make more sense to check if it has a parent instead? Using this implementation could cause products to be exported even though they shouldn't. Maybe I am missing something though.

TobiasGraml11: Because I only check the categories, that are really relevant to the product. But as we don't save the exported categories anywhere, I need to do this check here as well.

- If you configure a category as cross selling category, that is not in a different base category, but configured for this product, it should be exported. (e.g. The product is part of 2 sub-shops)
- There is a root category as a parent of the base categories, so we can't check for a valid parent. The string from the config input would never match.

TheKeymaster: Makes sense, since we are in the context of a product. Change lgtm.

PR #320, opened 2020-05-26
SW-445 Add count of zero to filter request

TobiasGraml11:


dblaichinger:
@TheKeymaster 
Thanks a lot for having a look, you make me more comfortable to make the release!

dblaichinger:
@TobiasGraml11 
>  please ensure that we didn't miss any other parameters than count and usergroup

In the screenshot you created with 9.2.0 the request with `count=0` also adds `first=0`, I think this makes no difference but can you please check if the response is different if not given.

TobiasGraml11:
@dblaichinger 
> In the screenshot you created with 9.2.0 the request with `count=0` also adds `first=0`, I think this makes no difference but can you please check if the response is different if not given

`first=0` does not change the response, as it is the default value. In 9.2.0 it was added, because every request was built in the same function. I could add this for consistency, but I think it is not necessary.

PR #324, opened 2020-06-23
SW-460 Fix usergroup hash in snippet

TobiasGraml11:


FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -195,7 +196,13 @@ public function onFrontendPostDispatch(Enlight_Event_EventArgs $args)
             return;
         }
 
-        $groupKey = Shopware()->Session()->get('sUserGroup', 'EK');
+        /** @var ContextService $context */
+        $context = Shopware()->Container()->get('shopware_storefront.context_service');
```
TheKeymaster, line 200: Please do a click-test for Shopware 5.2. It may be that this service didn't exist back then and then we must use the old implementation.

dblaichinger:
@TobiasGraml11 
Please always create a related story. I created one now: https://findologic.myjetbrains.com/youtrack/issue/SW-460.
Please update the description and use the id in the PR.

FinSearchUnified/Subscriber/Frontend.php
TobiasGraml11: Click tested in Shopware 5.2.26 without issues

FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -195,7 +196,13 @@ public function onFrontendPostDispatch(Enlight_Event_EventArgs $args)
...
+        $groupKey = $context->getShopContext()->getCurrentCustomerGroup()->getKey();
+        if (!$groupKey) {
```
TheKeymaster, line 202: That should not be needed. Shopware already defined this as fallback. See:

https://github.com/shopware/shopware/blob/v5.2.0/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/ContextService.php#L48

PR #325, opened 2020-07-09
SW-464 Export cat_url with base url

TobiasGraml11:


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -538,7 +538,9 @@ protected function setAttributes()
                 }
 
                 if (!StaticHelper::isEmpty($tempPath)) {
-                    $catUrlArray[] = $this->seoRouter->sCleanupPath($tempPath);
+                    $categoryPath = $this->seoRouter->sCleanupPath($tempPath);
+                    $catUrlArray[] = $categoryPath;
+                    $catUrlArray[] = Shopware()->Shop()->getBaseUrl() . $categoryPath;
```
TheKeymaster, line 543: How do we know that `Shopware()->Shop()` is the shop of the current service? E.g.

German Shop: `https://shop.com/de`. Configured Shopkey = `ABCD`
English Shop: `https://shop.de/en`. Configured Shopkey = `DCBA`

Running the export with `https://shop.com/de?shopkey=ABCD` gives the the German products with the German base URL, since I am in that store.
Running the export with `https://shop.com/de?shopkey=DCBA` gives the English products, but with the German base URL, since I am in that store.

----

In a real world the German shop may even be without base URL. In this case it may be valid to call the export just from the main URL, even from the English Store.

PR #328, opened 2020-09-17
SW-489_fix_sdym_dym_query

TheKeymaster:


FinSearchUnified/plugin.xml
```diff
@@ -62,11 +62,11 @@
     <changelog version="10.4.2">
         <changes lang="de">
             Die Basis-URL des Shops wurde bei cat_urls nicht berücksichtigt. Dies wurde behoben.;
-            Ein Fehler wurde behoben, welcher dazu führte, dass Filter nicht korrekt funktionierten, wenn diese multi-byte Zeichen beinhalteten.
+            Ein Fehler wurde behoben, welcher dazu führte, dass Filter nicht korrekt funktionierten, wenn diese multi-byte Zeichen beinhalteten.;
```
dblaichinger, line 65: Why dot and semicolon?

dblaichinger (changes requested):

FinSearchUnified/plugin.xml
TheKeymaster: Shopware adds `;` at the end of each line. Since I end my sentences with a dot, a dot is added before the line ending, which in that case is `;`.

dblaichinger (approved):

PR #331, opened 2020-11-04
SW-526 Export correct cat_url's

TobiasGraml11:


.travis.yml
```diff
@@ -68,11 +68,12 @@ jobs:
       before_install: skip
       before_script: skip
       script: composer lint
-    - stage: "Shopware 5.2.0"
-      php: "5.6"
-      env:
-        - SHOPWARE_TAG="5.2.0"
-        - SHOPWARE_VERSION="5.2"
+# Deactivated until the fix in SW-528
```
TheKeymaster, line 71: Please only deactivate the specific test case in case it is Shopware 5.2.0, and don't disable the entire test suite for it.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -530,15 +530,19 @@ protected function setAttributes()
...
                 $tempPath = sprintf($routerCategoryTemplate, implode('/', $catPath));
 
-                if (Shopware()->Config()->get('routerToLower')) {
+                if (Shopware()->Config()->offsetGet('routerToLower')) {
                     $tempPath = mb_strtolower($tempPath);
                 }
 
                 if (!StaticHelper::isEmpty($tempPath)) {
-                    $categoryPath = $this->seoRouter->sCleanupPath($tempPath);
+                    $categoryPath = '/' . ltrim($this->seoRouter->sCleanupPath($tempPath), '/');
```
TheKeymaster, line 545: Please add a comment above, why this is necessary, so this doesn't get removed in the future because someone thinks this is not necessary. I am also unsure about the last `/`, I may be wrong but I think there is a configuration, which allows  to change this.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -1320,11 +1332,117 @@ public function testMultiByteCharactersAreExportedInLowercase($articleConfigurat
...
+    /**
+     * @dataProvider categoriesWithSlashProvider
+     *
+     * @param array $articleConfiguration
+     *
+     * @throws Exception
+     */
+    public function testCategoryNamesWithSlashesAreExportedCorrectly($articleConfiguration)
+    {
+        $articleFromConfiguration = $this->createTestProduct($articleConfiguration);
```
TheKeymaster, line 1407: Use `$this->markTestSkipped()` with an appropriate message, in case it is Shopware 5.2. Use the environment for that, we already use it in other places in the tests.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #332, opened 2020-11-16
SW-434 add pseudoSale

KukinPavel:
Implementing "SW-434 As customer I want to use Shopware pseudo sales which should override the actual sales frequency in the export".

### DIFF
--- a/FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
+++ b/FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
@@ -470,8 +470,15 @@ class FindologicArticleModel
 
     protected function setSales()
     {
+        $articleSales = $this->getArticleSalesFrequency();
+        $salesFrequency = new SalesFrequency();
+        $salesFrequency->setValue($articleSales);
+        $this->xmlArticle->setSalesFrequency($salesFrequency);
+    }
+
+    protected function getArticleSalesFrequency() {
         $pseudoSales = $this->baseArticle->getPseudoSales();
-        if (is_int($pseudoSales) && $pseudoSales > 0) {
+        if ($pseudoSales !== null && $pseudoSales > 0) {
             $articleSales = $pseudoSales;
         } else {
             $orderDetailQuery = $this->orderDetailRepository->createQueryBuilder('order_details')
@@ -479,13 +486,9 @@ class FindologicArticleModel
                 ->where('order_details.articleId = :articleId')
                 ->select('sum(order_details.quantity)')
                 ->setParameter('articleId', $this->baseArticle->getId());
-
             $articleSales = (int)$orderDetailQuery->getQuery()->getScalarResult()[0][1];
         }
-
-        $salesFrequency = new SalesFrequency();
-        $salesFrequency->setValue($articleSales);
-        $this->xmlArticle->setSalesFrequency($salesFrequency);
+        return $articleSales;
     }
 
     protected function setAttributes()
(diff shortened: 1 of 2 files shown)


FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -470,13 +470,18 @@ protected function setImages()
 
     protected function setSales()
     {
-        $orderDetailQuery = $this->orderDetailRepository->createQueryBuilder('order_details')
-            ->groupBy('order_details.articleId')
-            ->where('order_details.articleId = :articleId')
-            ->select('sum(order_details.quantity)')
-            ->setParameter('articleId', $this->baseArticle->getId());
+        $pseudoSales = $this->baseArticle->getPseudoSales();
+        if (is_int($pseudoSales) && $pseudoSales > 0) {
```
TheKeymaster, line 474: As far as the docblock suggests, this may never return anything else than `int`:

image

In case you want to check for nullable, please be explicit by checking for `$pseudoSales !== null`.

FinSearchUnified/BusinessLogic/Models/FindologicArticleModel.php
```diff
@@ -470,13 +470,18 @@ protected function setImages()
...
+        } else {
+            $orderDetailQuery = $this->orderDetailRepository->createQueryBuilder('order_details')
+                ->groupBy('order_details.articleId')
+                ->where('order_details.articleId = :articleId')
+                ->select('sum(order_details.quantity)')
+                ->setParameter('articleId', $this->baseArticle->getId());
 
-        $articleSales = (int)$orderDetailQuery->getQuery()->getScalarResult()[0][1];
+            $articleSales = (int)$orderDetailQuery->getQuery()->getScalarResult()[0][1];
+        }
```
TheKeymaster, line 484: Move this into a separate method. e.g. `getSalesFrequency` or something similar.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -1453,4 +1453,81 @@ public function testCategoryNamesWithSlashesAreExportedCorrectly($articleConfigu
             $cat_urls
         );
     }
+
+    public function productsWithPseudoSalesProvider()
+    {
+        return [
+            'Pseudo sale value of 0 is not exported' => [
+                'pseudoSales' => 0,
+                'expectedSalesFrequency' => 1
```
TheKeymaster, line 1462: Names here should match the actual method parameter names. Either rename this here to `expected` or rename the method parameter to `$expectedSalesFrequency`.

TheKeymaster, line 1462: Where is `1` being set? I do not see any sales being created in the test.

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
```diff
@@ -1453,4 +1453,81 @@ public function testCategoryNamesWithSlashesAreExportedCorrectly($articleConfigu
...
+        ];
+    }
+
+    /**
+     * @dataProvider productsWithPseudoSalesProvider
+     *
+     * @param int $pseudoSales
+     * @param int $expected
+     *
+     * @throws Exception
```
TheKeymaster, line 1477: We do not really care about any `@throws` tags in tests, so this can be removed.

TheKeymaster (changes requested):

FinSearchUnified/Tests/BusinessLogic/Models/FindologicArticleModelTest.php
KukinPavel: In the test article that is created there are by default an order with one item with `quantity: 1` .

TheKeymaster (approved):

PR #333, opened 2021-01-12
SW-564 Make services public

TheKeymaster:


FinSearchUnified/Resources/services.xml
```diff
@@ -136,39 +166,61 @@
...
-        <service id="shopware.routing.matchers.rewrite_matcher"
-            class="Shopware\Components\Routing\Matchers\RewriteMatcher">
+        <service
+            id="shopware.routing.matchers.rewrite_matcher"
+            class="Shopware\Components\Routing\Matchers\RewriteMatcher"
+            public="true"
+        >
             <argument type="service" id="dbal_connection" />
             <argument type="service" id="query_alias_mapper" />
+            <argument type="service" id="Shopware_Components_Config"/>
```
dblaichinger, line 223: How did this work until now? Missing argument?

dblaichinger (changes requested):

FinSearchUnified/Resources/services.xml
TheKeymaster: In case the `RewriteMatcher` already exists, it will be called with the given arguments. Since Shopware added `Shopware_Components_Config` as new argument, we need to also add it https://github.com/shopware/shopware/commit/8a3470d7a39bb1853c1c8a368d0608aba691c8ee. Since the build broke anyway, I made this ignore on invalid.

dblaichinger (approved):

PR #334, opened 2021-01-13
SW-563_fix_image_filters

TheKeymaster:


FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
```diff
@@ -367,12 +372,22 @@ public function facetWithVendorFilterProvider()
                     true,
                     'Hersteller',
                     [
-                        new MediaListItem('Anderson, Gusikowski and Barton', 'Anderson, Gusikowski and Barton', true),
-                        new MediaListItem('Bednar Ltd', 'Bednar Ltd', false)
+                        new MediaListItem(
+                            'Anderson, Gusikowski and Barton',
```
dblaichinger, line 376: Is the name also the id?

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/ImageFacetHandlerTest.php
```diff
@@ -68,48 +68,42 @@ public function imageFilterProvider()
...
+                    Operator::EQ,
+                    ['Red', 'Zima Blue', 'Purple']
+                ),
+                'facetData' => [
+                    ['id' => 'Red', 'active' => true, 'media' => null],
+                    ['id' => 'Green', 'active' => false, 'media' => null],
+                    ['id' => 'Zima Blue', 'active' => true, 'media' => 'https://example.com/zima-blue.gif'],
+                    ['id' => 'Yellow', 'active' => false, 'media' => 'https://example.com/yellow.gif'],
+                    ['id' => 'Purple', 'active' => true, 'media' => 'https://example.com/purple.gif'],
+                    ['id' => 'Light Purple', 'active' => false, 'media' => 'https://example.com/light-purple.gif'],
```
dblaichinger, line 82: Why is there a media url and before it wasn't?

dblaichinger (changes requested):

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/FacetHandler/ImageFacetHandlerTest.php
TheKeymaster: Because before this would be `null` if Guzzle wasn't able to receive the image. In the new implementation we're always setting an image.

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
TheKeymaster: Yes Shopware works with ids. The only difference would be special characters, which we have tests for. The ID can only contain data that can be written in an HTML name attribute.

FinSearchUnified/Resources/views/frontend/less/color-filter.less
```diff
@@ -1,6 +1,10 @@
-.color--filter-item {
-  width: 30px;
-  height: 30px;
+.input-type--media .option--container .filter-panel--label {
+  width: 45px;
+  height: 45px;
   border: 1px solid #dbd7e5;
-  margin: 2px;
+
+  div {
```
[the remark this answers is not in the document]
TheKeymaster: You're right, I've added an additional class selector.

FinSearchUnified/Tests/Bundle/ProductNumberSearchTest.php
TheKeymaster: This is from the HTML source:

image

The id represents the name as escaped data and the name represents the actual shown value. If I recall correctly in Shopware 5, this will be done automatically by Shopware.

dblaichinger (approved):

PR #336, opened 2021-01-14
SW-568_switch_to_github_actions

TheKeymaster:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
...
+          composer validate
+
+      - name: Install dependencies
+        run: |
+          cd FinSearchUnified
+          composer install
+
+      - name: Run code linting
+        run: |
+          cd FinSearchUnified
```
dblaichinger, line 43: Does worklflows allow to set a default directory to avoid those `cd`'s?

.github/workflows/phpunit_sw_old_versions.yml
```diff
@@ -0,0 +1,88 @@
...
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: ['5.6', '7.0', '7.1']
+        shopware-versions: ['5.2', 'v5.2.0', '5.3', 'v5.3.0', '5.4', '5.5']
```
dblaichinger, line 18: Why do we test 5.2. and 5.3. twice?

dblaichinger (changes requested):

.github/workflows/phpunit_sw_old_versions.yml
TheKeymaster: Because 5.2 = 5.2.27 and 5.3 = 5.3.6. The other two versions are explicitly pinned, so ensure that our plugin also works with these .0 versions, as we had issues with them im the past.

.github/workflows/lint.yml
TheKeymaster: Since we have Shopware and our plugin as separate Git repositories, we manually need to go into these directories. Also our plugin will be moved inside the Shopware's plugin directory before installing.
It is also way easier to work with, as you do not need to know where you're currently are, because you will always be in the root folder at the beginning of each task, which really helped a lot.

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -363,7 +364,7 @@ public static function setQueryInfoMessage(QueryInfoMessage $queryInfoMessage)
      */
     public static function isVersionLowerThan($version)
     {
-        $shopwareVersion = getenv('SHOPWARE_VERSION') ?: '5.6.7';
+        $shopwareVersion = static::getShopwareVersion();
```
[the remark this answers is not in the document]
TheKeymaster: Shopware's linter is actually strict about this and suggests avoiding the use of `self`, as it is discouraged. See this screenshot from a while ago:

image

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -373,4 +374,20 @@ public static function isVersionLowerThan($version)
...
+    {
+        $version = null;
+        if (Shopware()->Container()->has('shopware.release.version')) {
+            $version = Shopware()->Container()->get('shopware.release.version');
+        } elseif (defined('\Shopware::VERSION')) {
+            $version = Shopware::VERSION;
+        }
+
+        if (!$version || $version === '___VERSION___') {
+            $version = getenv('SHOPWARE_VERSION') ?: '5.6.7';
```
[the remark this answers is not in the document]
TheKeymaster: This was here before, but I guess if Shopware returns `___VERSION___`, we would have a problem if `SHOPWARE_VERSION` also isn't set in the environment, because then we wouldn't have a version at all, which is bad. This doesn't mean that I am suggesting hard coding a random version is better than no version at all.

.github/workflows/phpunit_sw_old_versions.yml
dblaichinger: Ok just saw we did the same in Travis, resolved.

FinSearchUnified/Helper/StaticHelper.php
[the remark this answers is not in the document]
dblaichinger: Should the default be the latest version? Currently it would be 5.6.9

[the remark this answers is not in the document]
TheKeymaster: Sure, we can do that. Done.

dblaichinger (approved):

PR #339, opened 2021-02-02
SW-577_fix_encoding

TheKeymaster:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/ConditionHandler/ProductAttributeConditionHandlerTest.php
```diff
@@ -70,10 +71,10 @@ public function attributesDataProvider()
                 ],
                 ['color' => ['' => ['blue', 'red']]]
             ],
-            'Vendor is "Brand+Name" and color is "red"' => [
+            'Vendor is "Brand Name" and color is "red"' => [
                 Operator::EQ,
                 [
-                    ['field' => 'vendor', 'value' => 'Brand+Name'],
```
dblaichinger, line 76: Was the this case testing the wrong thing? When encoded twice this would result in `Brand%2BName`.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/ConditionHandler/ProductAttributeConditionHandlerTest.php
```diff
@@ -143,4 +144,33 @@ public function testGenerateCondition($operator, array $attributes, array $expec
...
+            new ProductAttributeCondition($filterName, Operator::EQ, $filterValue),
+            $this->querybuilder,
+            $this->context
+        );
+
+        $params = $this->querybuilder->getParameters();
+
+        static::assertArrayHasKey('attrib', $params);
+        static::assertArrayHasKey($filterName, $params['attrib']);
+        static::assertSame($filterValue, $params['attrib'][$filterName]['']);
```
dblaichinger, line 169: Why is there `['']`?

dblaichinger (changes requested):

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/ConditionHandler/ProductAttributeConditionHandlerTest.php
TheKeymaster: It is more like this:
1. User submits filter. URL contains `...filterName=Brand%2BName`
2. When Shopware builds their `Criteria` event, their `Request` class already contains all query related information URL decoded, therefore this value then is `Brand+Name`. This is already done at a PHP level, where PHP populates the superglobal `$_GET` with query information which is **already url decoded**.
3. In our old logic we then took `Brand+Name` and url decoded it again, so it will be `Brand Name`, which is a completely different filter.

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/ConditionHandler/ProductAttributeConditionHandlerTest.php
TheKeymaster: This is due to how the URL structure of Findologic is. It is always `attrib[filterName][]=filterValue`, therefore `''` is just accessing that empty string array key. For a range slide for example it would be `attrib[filterName][min/max]=filterValue` and instead of `''` you would get the data from `'min'`/`'max'`.

dblaichinger (approved):

PR #340, opened 2021-03-25
SW-597 Add general exception handling for products

TobiasGraml11:


FinSearchUnified/BusinessLogic/ExportService.php
```diff
@@ -231,6 +231,17 @@ public function getFindologicArticle($shopwareArticle)
                     $shopwareArticle->getMainDetail()->getNumber()
                 )
             );
+        } catch (EntityNotFoundException $e) {
```
TheKeymaster, line 234: Why aren't we just catching `\Exception` here instead? We do this for Shopware 6, but with the difference that we explicitly add `... If you see this message in your logs, please report this as a bug ...`:

https://github.com/findologic/plugin-shopware-6/blob/0dd54ada1ff1d46ab3febde8e52628b2fc0f9957/src/Export/XmlProduct.php#L258

In Shopware 6 we use `\Throwable`, but we can not do this for Shopware 5, as Shopware 5.2 and I also think 5.3 still support PHP 5.6.

FinSearchUnified/Tests/PluginTest.php
```diff
@@ -240,4 +241,19 @@ public function testEmptyValueNotAllowedExceptionIsThrownInExport()
         $exported = Utility::runExportAndReturnCount();
         $this->assertSame(0, $exported);
     }
+
+    public function testEntityNotFoundExceptionIsThrownInExport()
```
TheKeymaster, line 245: Please also ensure that the error message is logged.

TheKeymaster (changes requested):

FinSearchUnified/BusinessLogic/ExportService.php
dblaichinger: @TheKeymaster The suggestion to just catch `EntityNotFoundException` came from me. I thought it's maybe not clear from the customer-perspective when complete products are missing because we skip it always. But if we already use the same logic for SW6, I'll guess it's fine for SW5 as well.

TheKeymaster: We can also separate them and add an additional catch for the `EntityNotFoundException`.

```php
// ...
} catch (EntityNotFoundException $e) {
    // ... log that the entity was not found somehow.
} catch (\Exception $e) {
   // ... log that something went wrong.
}
```

FinSearchUnified/BusinessLogic/ExportService.php
```diff
@@ -223,14 +223,37 @@ public function getFindologicArticle($shopwareArticle)
...
+            Shopware()->Container()->get('pluginlogger')->error(
                 sprintf(
                     'Product with number "%s" could not be exported. ' .
                     'It appears to have empty values assigned to it. ' .
                     'If you see this message in your logs, please report this as a bug',
                     $shopwareArticle->getMainDetail()->getNumber()
                 )
             );
+        } catch (EntityNotFoundException $e) {
+            $errorInformation->addError('EntityNotFoundException' . $e->getMessage());
```
TheKeymaster, line 235: Add a space, as otherwise this may happen:

```text
EntityNotFoundExceptionArticle could not be found.
```

```suggestion
            $errorInformation->addError('EntityNotFoundException ' . $e->getMessage());
```

FinSearchUnified/BusinessLogic/ExportService.php
```diff
@@ -223,14 +223,37 @@ public function getFindologicArticle($shopwareArticle)
...
+                sprintf(
+                    'Product with ID "%s" could not be exported. ' .
+                    'It appears to have a non existing variant configured in the database. ' .
+                    'Error message: %s',
+                    $shopwareArticle->getId(),
+                    $e->getMessage()
+                )
+            );
+        } catch (Exception $e) {
+            $errorInformation->addError('Exception' . $e->getMessage());
```
TheKeymaster, line 247:
```suggestion
            $errorInformation->addError('Exception: ' . $e->getMessage());
```

FinSearchUnified/Tests/PluginTest.php
```diff
@@ -226,15 +229,39 @@ public function testArticleExportWithCrossSellingCategories($crossSellingCategor
         $this->assertSame($expectedCount, $exportedCount);
     }
 
-    public function testEmptyValueNotAllowedExceptionIsThrownInExport()
+    public function exceptionProvider()
+    {
+        return [
+            'EmptyValueNotAllowedException' => [
+                'exception' =>  new EmptyValueNotAllowedException(),
```
TheKeymaster, line 236: Add a `message` argument as well and ensure that the log message contains the given message.
```suggestion
                'exception' =>  new EmptyValueNotAllowedException('Entity not found'),
                'expectedMessage' => 'Entity not found',
```

FinSearchUnified/Tests/PluginTest.php
```diff
@@ -226,15 +229,39 @@ public function testArticleExportWithCrossSellingCategories($crossSellingCategor
...
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider exceptionProvider
+     *
+     * @param $exception
+     */
+    public function testExceptionsAreThrown($exception)
```
TheKeymaster, line 252: This can be typehinted. With a typehint you can remove the parameter in the docblock.
```suggestion
    public function testExceptionsAreThrown(Exception $exception)
```

TheKeymaster (changes requested):

FinSearchUnified/Tests/PluginTest.php
TobiasGraml11: I can't find a way to do that, as `willThrowException` is the only exception-related assertion I could find for the mock.

TheKeymaster: Yeah here it is kinda obvious, as you're telling the mock to throw the exception. But I want to ensure that the logged message contains the exception message.

```php
// OLD
$loggerMock->expects($this->once())
    ->method('error');

// NEW
$loggerMock->expects($this->once())
    ->method('error')
    ->with($expectedExceptionMessage);
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #341, opened 2021-04-08
SW-598 Configure Cross Selling category for each shop

TobiasGraml11:


FinSearchUnified/Tests/PluginTest.php
```diff
@@ -190,37 +191,74 @@ public function testProductIdExport($productId, $errorMessageOrProductCount)
...
-     * @param int[] $crossSellingCategories
+     * @param int[] $baseCrossSellingCategories
      * @param int $expectedCount
+     * @param Shop $baseShop,
+     * @param ?Shop $subShop,
+     * @param int[] $subCrossSellingCategories,
      */
-    public function testArticleExportWithCrossSellingCategories($crossSellingCategories, $expectedCount)
-    {
+    public function testArticleExportWithCrossSellingCategories(
```
TheKeymaster, line 244: Does this fail if you remove `scope="shop"` from `FinSearchUnified/Resources/config/config_template.xml`? What's the error message?

TheKeymaster (changes requested):

FinSearchUnified/Tests/PluginTest.php
TobiasGraml11: It does not seem to fail. As we need to set the config manually, I am not sure if we can check that.

TheKeymaster: @dblaichinger and I have discussed this and we will leave this as it is, even though it does not explicitly test that the `scope` setting is checked, but we also couldn't really find any better alternatives.

TheKeymaster (approved):

PR #344, opened 2021-07-14
SW-626 Decode HTML code

TobiasGraml11:


FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -362,6 +362,11 @@ public static function cleanStringProvider()
                 'Findolögic123',
                 'Findolögic123',
                 'Expected umlauts to be left unaltered.'
+            ],
+            'String with html code' => [
+                'With&nbsp;HTML&nbsp;code',
+                html_entity_decode('With&nbsp;HTML&nbsp;code'),
```
TheKeymaster, line 368: `&nbsp;` is such an odd character, as the reason why you can not really type `With HTML code` here, is that `&nbsp;` results in white-space, which would URL encoded look like `%C2%A0`. We should know what the result is, so maybe we can use different HTML characters such as `&amp;` or `&gt;`. This way we can ensure that HTML entities are encoded, without having to call `html_entity_decode` inside of the tests.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #345, opened 2021-07-20
SW-628 Support pushAttrib in search form

TobiasGraml11:


FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder/QueryBuilderFactory.php
```diff
@@ -148,6 +149,20 @@ private function addSorting(Criteria $criteria, QueryBuilder $query, ShopContext
         }
     }
 
+    /**
+     * @param QueryBuilder $query
+     */
+    private function addPushAttribs(QueryBuilder $query)
```
TheKeymaster, line 155: As we still support PHP 5.6, please remove the typehint. In the docblock it should still stay though.

FinSearchUnified/Resources/views/frontend/listing/actions/action-filter-panel.tpl
```diff
@@ -0,0 +1,13 @@
+{extends file="parent:frontend/listing/actions/action-filter-panel.tpl"}
+
+{block name="frontend_listing_actions_filter_form_page"}
+    {if $finPushAttribs}
+        {foreach $finPushAttribs as $finFilterName => $finFilterValues}
+            {foreach $finFilterValues as $finFilterValue => $finWeight}
+                <input type="hidden" name="pushAttrib[{$finFilterName}][{$finFilterValue}]" value="{$finWeight}"/>
```
TheKeymaster, line 7: This may allow XSS in some form or another. Please sanitize all three variables before setting them here.

TheKeymaster (changes requested):

FinSearchUnified/Bundle/SearchBundleFindologic/QueryBuilder/QueryBuilderFactory.php
TobiasGraml11: Aren't typehints of classes supported in PHP 5.6?
https://mlocati.github.io/articles/php-type-hinting.html

TheKeymaster: You are of course totally correct. My bad, haven't used such a legacy version in a while. Was just thinking about it and that's why I commented.

TheKeymaster (approved):

PR #346, opened 2021-08-04
SW-629 Fix channels without a succesfull import

TobiasGraml11:


TobiasGraml11:
Closed, because it was fixed by SW 5.7

PR #348, opened 2021-08-18
SW-640 Add custom color filter images

TobiasGraml11:


FinSearchUnified/Bundle/SearchBundle/FacetResult/ColorListItem.php
```diff
@@ -11,17 +12,24 @@ class ColorListItem extends FacetResult\MediaListItem
      */
     protected $colorcode;
 
+    /**
+     * @var string
```
TheKeymaster, line 16:
```suggestion
     * @var string|null
```

FinSearchUnified/Bundle/SearchBundle/FacetResult/ColorListItem.php
```diff
@@ -39,4 +47,12 @@ public function setColorcode($colorcode)
     {
         $this->colorcode = $colorcode;
     }
+
+    /**
+     * @return string
```
TheKeymaster, line 52:
```suggestion
     * @return string|null
```

FinSearchUnified/Resources/views/frontend/listing/filter/_includes/filter-color-selection.tpl
```diff
@@ -21,9 +21,13 @@
     <label class="filter-panel--label"
            for="__{$facet->getFieldName()|escape:'htmlall'}__{$option->getId()|escape:'htmlall'}">
 
-        {if $option->getColorcode()}
-            {$mediaFile = $option->getColorcode()}
-            <div class="color--filter-item" style="background: {$mediaFile};" title="{$option->getId()|escape:'htmlall'}"></div>
+        {$colorCode = $option->getColorcode()}
+        {$imageUrl = $option->getImageUrl()}
+
+        {if $colorCode && $imageUrl}
```
TheKeymaster, line 27: Shouldn't checking for `$imageUrl` be sufficient? We are we also checking for `$colorCode`?

FinSearchUnified/Resources/views/frontend/listing/filter/_includes/filter-color-selection.tpl
```diff
@@ -21,9 +21,13 @@
...
+            <div class="color--filter-item" style="background-image: url({$imageUrl}); background-color: {$colorCode};" title="{$option->getId()|escape:'htmlall'}"></div>
+        {elseif $colorCode}
+            <div class="color--filter-item" style="background: {$colorCode};" title="{$option->getId()|escape:'htmlall'}"></div>
         {else}
```
TheKeymaster, line 31: Wouldn't this mean that if an image does not have a color code, but has a custom `$imageUrl`, that always the no-picture is shown?
In case it is not, those conditions are not really self-explanatory, can we make this a bit more clear?

TheKeymaster (changes requested):

FinSearchUnified/Resources/views/frontend/listing/filter/_includes/filter-color-selection.tpl
TobiasGraml11: I added a check for `$imageUrl` separately, so we can provide the no-result image as backup

FinSearchUnified/Resources/views/frontend/listing/filter/_includes/filter-color-selection.tpl
TobiasGraml11: I added another if statement for `$imageUrl`. Should be more explanatory now.

TheKeymaster (approved):

PR #350, opened 2021-10-28
SW-655 Use correct product_number_search service

TobiasGraml11:


FinSearchUnified/FinSearchUnified.php
```diff
@@ -34,14 +39,15 @@ public function build(ContainerBuilder $container)
...
-            );
-
             $loader->load($this->getPath() . '/Resources/shopware/searchBundleES.xml');
         }
 
+        if ($container->getParameter('shopware.es.enabled')) {
+            $loader->load($this->getPath() . '/Resources/shopware/productSearch.xml');
+        } else {
+            $loader->load($this->getPath() . '/Resources/shopware/productSearch.xml');
+        }
```
TheKeymaster: Shouldn't the first one be `productSearchES.xml`?
```suggestion
        if ($container->getParameter('shopware.es.enabled')) {
            $loader->load($this->getPath() . '/Resources/shopware/productSearchES.xml');
        } else {
            $loader->load($this->getPath() . '/Resources/shopware/productSearch.xml');
        }
```

TheKeymaster (changes requested):

FinSearchUnified/FinSearchUnified.php
TobiasGraml11: Oh, missed that while testing locally

TheKeymaster (approved):

PR #357, opened 2022-05-09
SW-602 include shop system and version in the findologic request

nenadj-soprex:


FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilder/SearchQueryBuilderTest.php
```diff
@@ -380,4 +380,38 @@ public function testForceOriginalQueryIsNotTreatedAsFilter()
...
+            null,
+            $searchNavigationRequest
+        );
+
+        $queryBuilder->addQuery('shopware5');
+
+        $params = $searchNavigationRequest->getParams();
+        $this->assertArrayHasKey('query', $params);
+        $this->assertSame('shopware5', $params['query']);
+    }
```
TobiasGraml11: This also applies to the other test:
- You are currently not testing the correct parameter, as you just set the query and test it
- As the parameter should be added by default, you don't need any additional call to e.g. `addQuery`
- It's usually a good idea to comment out your changes, and see if the test fails without it. This is not the case here

```suggestion
    public function testShopTypeParameterIsAddedByDefault()
    {
        $searchNavigationRequest = new SearchRequest();
        new SearchQueryBuilder(
            $this->installerService,
            $this->config,
            null,
            $searchNavigationRequest
        );

        $params = $searchNavigationRequest->getParams();
        $this->assertArrayHasKey('shopType', $params);
        $this->assertSame('Shopware5', $params['shopType']);
    }
```

TobiasGraml11 (changes requested):

FinSearchUnified/Tests/Bundle/SearchBundleFindologic/QueryBuilder/SearchQueryBuilderTest.php
nenadj-soprex: Thanks for the tip, we implemented the solution and now it seems ok.

TobiasGraml11 (approved):

PR #360, opened 2022-05-19
SW-725 Ajax requests for manufacturer pages must not be handled in our plugin

nenadj-soprex:


FinSearchUnified/Subscriber/Frontend.php
```diff
@@ -106,12 +106,15 @@ public function onFrontendPreDispatch(Enlight_Event_EventArgs $args)
...
+            Shopware()->Session()->offsetSet('isManufacturerPage', false);
         } elseif (StaticHelper::isCategoryPage($request)) {
             Shopware()->Session()->offsetSet('isCategoryPage', true);
             Shopware()->Session()->offsetSet('isSearchPage', false);
-        } elseif ($this->isManufacturerPage($request) || !$this->isWhiteListed($request)) {
+            Shopware()->Session()->offsetSet('isManufacturerPage', false);
+        } elseif (StaticHelper::isManufacturerPage($request) || !$this->isWhiteListed($request)) {
             Shopware()->Session()->offsetSet('isCategoryPage', false);
             Shopware()->Session()->offsetSet('isSearchPage', false);
+            Shopware()->Session()->offsetSet('isManufacturerPage', StaticHelper::isManufacturerPage($request));
```
TobiasGraml11: Just realised, that we will need to do this:
```suggestion
        } elseif (StaticHelper::isManufacturerPage($request)) {
            Shopware()->Session()->offsetSet('isCategoryPage', false);
            Shopware()->Session()->offsetSet('isSearchPage', false);
            Shopware()->Session()->offsetSet('isManufacturerPage', true);
        } elseif (!$this->isWhiteListed($request)) {
            Shopware()->Session()->offsetSet('isCategoryPage', false);
            Shopware()->Session()->offsetSet('isSearchPage', false);
            Shopware()->Session()->offsetSet('isManufacturerPage', false);
```

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -455,6 +455,38 @@ public function testUseShopSearch(
...
+
+        // Assign mocked config variable to application container
+        Shopware()->Container()->set('config', $config);
+
+        Shopware()->Session()->offsetSet('isManufacturerPage', false);
+
+        $result = StaticHelper::useShopSearch();
+        $this->assertTrue($result, 'Expected Shopware search to be triggered but it was not');
+    }
+
```
TobiasGraml11: Should be enough to just extend `testUseShopSearch` and its data provider `shopSearchProvider`

FinSearchUnified/Subscriber/Frontend.php
TobiasGraml11: We should catch this via test as well

TobiasGraml11 (changes requested):

FinSearchUnified/Helper/StaticHelper.php
```diff
@@ -193,6 +196,17 @@ public static function isCategoryPage(Enlight_Controller_Request_Request $reques
...
 
+    /**
+     * @param Enlight_Controller_Request_Request $request
+     *
+     * @return bool
+     */
+    public static function isManufacturerPage(Enlight_Controller_Request_Request $request)
+    {
+        return $request->getControllerName() === 'listing' &&
+            ($request->getActionName() === 'manufacturer' || $request->getActionName() === 'listingCount');
```
TobiasGraml11: This is not the exact same as the one in `Frontend`. Why was the `$request->getActionName() === 'listingCount'` part added?

FinSearchUnified/Tests/Helper/StaticHelperTest.php
```diff
@@ -223,16 +230,29 @@ public static function shopSearchProvider()
...
+                'activateFindologic' => true,
+                'shopKey' => 'ABCDABCDABCDABCDABCDABCDABCDABCD',
+                'activateFindologicForCategoryPages' => true,
+                'findologicDI' => false,
+                'isSearchPage' => false,
+                'isCategoryPage' => false,
+                'isManufacturerPage' => true,
+                'fallbackSearchCookie' => null,
+                'expected' => true
+            ],
```
TobiasGraml11: This does not fail, when removing the fix, because `$isNoSearchAndCategoryPage` will be true.
I suggest to do `'isCategoryPage' => true`.

FinSearchUnified/Helper/StaticHelper.php
nenadj-soprex: Sorry for that. Because when I worked on that task, on Manufacturer page, initial request was handled by 'manufacturer' action, but when we change the sort, that request was handled by 'listingCount' action.
I try today to reproduce this case but now every request handled by 'manufacturer' action. I will remove this additional condition in next commit.

FinSearchUnified/Tests/Helper/StaticHelperTest.php
nenadj-soprex: Thanks, we tested it, with and without fix. Now it works as expected.

TobiasGraml11 (approved):

PR #362, opened 2022-05-23
SW-723 the multi select color filter must work for live filter reloading

nenadj-soprex:


TobiasGraml11:
Moved to will-not-fix

